PR #12, opened 2016-11-25
Add travis config and clean up to coding standards

whikloj:
Related to https://github.com/Islandora-CLAW/CLAW/issues/435
This is big, but most is whitespace and punctuation.

I added the line_endings.sh and travis-setup.sh scripts, but we are not using travis-setup.sh right now. If we get some future Drupal tests in here, we could run them possibly just with phpunit.

This will get line endings, drupal coding standards and copy/paste detection started.


dannylamb:
So is @file no longer required?

whikloj:
@dannylamb 
"The @file doc block MUST be present for all PHP files, with one exception: files that contain a namespaced class/interface/trait, whose file name is the class name with a .php extension, and whose file path is closely related to the namespace (under PSR-4 or a similar standard), SHOULD NOT have a @file documentation block."

dannylamb:
Ok, so basically functional files like d7 still need them.

whikloj:
Yes, so I think a `.module` file or `.inc` file should have this annotation because it might be less clear what the contents are for.

dannylamb:
@whikloj I'm getting this when doing a `drupal cache:rebuild all` after pulling down the changes:

```
PHP Warning:  Declaration of Drupal\islandora\Form\FedoraResourceEntityInlineForm::submitCleanFormState(array &$entity_form, Drupal\Core\Form\FormStateInterface $form_state) should be compatible with Drupal\inline_entity_form\Form\EntityInlineForm::submitCleanFormState(&$entity_form, Drupal\Core\Form\FormStateInterface $form_state) in /var/www/html/drupal/modules/islandora/src/Form/FedoraResourceEntityInlineForm.php on line 21
```

src/Form/FedoraResourceEntityInlineForm.php
```diff
@@ -232,14 +238,15 @@ protected function buildEntity(array $entity_form, ContentEntityInterface $entit
...
    *
-   * @param $entity_form
+   * @param array $entity_form
    *   The entity form.
-   * @param $form_state
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
    *   The form state of the parent form.
    */
-  public static function submitCleanFormState(&$entity_form, FormStateInterface $form_state) {
+  public static function submitCleanFormState(array &$entity_form, FormStateInterface $form_state) {
```
dannylamb: Looks like the inline_entity_form modue is not 'in line' with Drupal coding standards.  You'll need to pull this array type hint until they fix it.

whikloj:
😮  *Gah*

dannylamb:
I know, right?  I guess technically we have to tell it to ignore standards, which iirc was always a bit wonky with code sniffer.

whikloj:
@dannylamb removed the type hint.

PR #13, opened 2016-11-25
Issue 430

Natkeeran:
This is PR for [issue#430](https://github.com/Islandora-CLAW/CLAW/issues/430).  

Notes
* The custom fields are by default disabled.  Should they be enabled by default?
* Both fields are of type string.  Please provide input as to what type they should be.
* (@dannylamb.  REST UI seems to work fine with the latest.  I suppose we don't need to set drupal permissions related to them anymore.) Currently for rdf mapping, description is mapped to dc:description, and memberOf is mapped to dc:isPartOf.  These will change based on the data model.

```
    {
        "@context":
        {
            "ldp": "http://www.w3.org/ns/ldp#",
            "dc11": "http://purl.org/dc/elements/1.1/",
            "nfo": "http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/v1.1/",
            "ebucore": "http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#",
            "fedora": "http://fedora.info/definitions/v4/repository#",
            "owl": "http://www.w3.org/2002/07/owl#",
            "ore": "http://www.openarchives.org/ore/terms/",
            "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
            "content": "http://purl.org/rss/1.0/modules/content/",
            "dc": "http://purl.org/dc/terms/",
            "foaf": "http://xmlns.com/foaf/0.1/",
            "og": "http://ogp.me/ns#",
            "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
            "schema": "http://schema.org/",
            "sioc": "http://rdfs.org/sioc/ns#",
            "sioct": "http://rdfs.org/sioc/types#",
            "skos": "http://www.w3.org/2004/02/skos/core#",
            "xsd": "http://www.w3.org/2001/XMLSchema#"
        },
        "@graph":
        [
            {
                "@id": "http://localhost:8000/fedora_resource/2?_format=jsonld",
                "@type":
                [
                    "schema:Thing",
                    "ldp:RDFSource",
                    "ldp:Container"
                ],
                "dc:title":
                [
                    {
                        "@value": "testabc"
                    }
                ],
                "rdf:label":
                [
                    {
                        "@value": "testabc"
                    }
                ],
                "schema:dateCreated":
                [
                    {
                        "@value": "2016-11-25T20:50:13+00:00"
                    }
                ],
                "dc:created":
                [
                    {
                        "@value": "2016-11-25T20:50:13+00:00"
                    }
                ],
                "schema:dateModified":
                [
                    {
                        "@value": "2016-11-25T20:51:33+00:00"
                    }
                ],
                "dc:description":
                [
                    {
                        "@value": "test description"
                    }
                ],
                "dc:isPartOf":
                [
                    {
                        "@value": "BasicImage"
                    }
                ]
            }
        ]
    }

```


dannylamb:
@Natkeeran thx so much for this.  I'll pull it down and see how it goes.

islandora_collection/config/install/field.field.fedora_resource.collection.field_description.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
```
dannylamb: Add and `enforced` entry here in order to get the configs to remove themselves when the module is uninstalled.  All the config files will need something like this.  You new `dependencies` for this file would look like:

```yaml
dependencies:
  config:
    - field.storage.fedora_resource.field_description
    - islandora.fedora_resource_type.collection
  enforced:
    module:
      - islandora_collection
```

dannylamb:
I think the conclusion from https://github.com/Islandora-CLAW/CLAW/issues/430 was to put `islandora_collection` nested inside a `modules` folder, but looking at this and `devel` makes me think what you've done is correct, @Natkeeran.  @whikloj thoughts?

Personally, I have no preference, but would like to do what is 'best practice', if that's possible/exists.  Initial google searches for 'drupal 8 sub module' aren't really coming up with anything decisive.

Natkeeran:
@dannylamb, Yes, putting the sub modules into its own folder would be the clean. 

```
    CLAW/islandora
    CLAW/islandora/modules
    CLAW/islandora/modules/islandora_collection
```
I can re-org.  Would like @whikloj to confirm it.

islandora_collection/config/install/field.storage.fedora_resource.field_description.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - islandora
+id: fedora_resource.field_description
+field_name: field_description
+entity_type: fedora_resource
+type: string
```
dannylamb: We'll want this to be `text` instead of `string` so we can have more characters and show trimmed versions or summaries for search results.

When you go to create the field, select `Text (formatted, long, with summary)` from the drop down.

islandora_collection/config/install/field.storage.fedora_resource.field_memberof.yml
```diff
@@ -0,0 +1,21 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - islandora
+id: fedora_resource.field_memberof
+field_name: field_memberof
+entity_type: fedora_resource
+type: string
```
dannylamb: We'll want this to be an entity reference field instead of `string` so we can render it inline and stuff.

You'll want to choose `Fedora resource` under the `Reference` category from the dropdown to create the field. Restrict bundle types to `Collection` for now.

Let me know if you need more help setting this up.

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
```diff
@@ -0,0 +1,46 @@
+langcode: en
```
dannylamb: This brings more questions to light about metadata modeling and what should be default.  It's functioning, so I don't want the mapping to block merging this PR.  But I think this warrants a second ticket to review so we an settle on a default.

DiegoPino:
I personally would like to avoid sub-anything if possible. Submodules on Drupal 8 are not standard anymore and we already had some issues updating, deploying with github sub..  not sure, i will of course we fine with whatever you decide.

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
dannylamb: https://github.com/Islandora-CLAW/CLAW/issues/439

dannylamb:
@DiegoPino @ruebot Not git submodules.  Drupal sub modules.  I don't want anymore Git submodules.

dannylamb:
Ninja'd by @ruebot

DiegoPino:
@ruebot i got that. Submodules in Drupal 8 have issues, because name spaces in classes in php depend the root resource. So it really really gets messy. If anyone can share a working, best practice example, i'm fine on that, but i know there are not and forums(and practice) point to not using it for the sake of being able to extend and use namespace -> module name naming conventions for classes. 
https://www.drupal.org/docs/develop/coding-standards/psr-4-namespaces-and-autoloading-in-drupal-8

dannylamb:
@DiegoPino FWIW I was looking at devel.  And yes, I came up short finding any sort of best practice.  Sounds like you're suggesting what I initially did, which would be one git repo, with each module as it's own top-level folder?

dannylamb:
@ruebot Whaaaaaa?  You don't like monotony?

DiegoPino:
@ruebot in that case i'm 80% clueless, how to we manage class autoloading?

dannylamb:
@ruebot Is putting everything in top level folders acceptable to get on drupal.org?  I'll have to do reserach.

Like, if the `islandora` git repo looked like
- ./islandora
- ./islandora_collection
- ./islandora_image (i'm just making this up)
- ...etc

Seems like that should appease both you and @DiegoPino, and would be clean from a repo point of view too.

dannylamb:
So are we cool with having @Natkeeran move forward with that repo structure above?

Thx for being patient while we sort this out, Nat.

whikloj:
I'm cool with that.

Natkeeran:
@dannylamb 

As I understand it, I can leave the islandora_collecition folder/module where it is right now (directly under https://github.com/Islandora-CLAW/islandora)!

I have to make the changes you've outlined above with respect to configs.

dannylamb:
@Natkeeran You are correct.  After https://github.com/Islandora-CLAW/islandora/pull/12 lands, I'll make a separate ticket for nesting islandora in its own separate folder.

dannylamb:
#12 has landed, as well as #14.  You'll want to pull those changes into your local branch.

Natkeeran:
@dannylamb Please review and let me know if commits look ok.  Thank you.

dannylamb:
Awesome.  Just gave it a whirl and it works as expected.  Configs are removing themselves upon uninstall, and fields function as I would expect in display and form display.

@Natkeeran Your contribution is greatly appreciated.

Natkeeran:
@dannylamb @ruebot Thank you.  Good to contribute something tangible.

Natkeeran:
@DiegoPino Please review the rdf mapping related discussions here: https://github.com/Islandora-CLAW/islandora/pull/15

The initial rdf mapping was a first first go.  @dannylamb has an updated version: https://github.com/Islandora-CLAW/islandora/pull/15/commits/776de8e6223a7e756d89c00789a219cdbc39e40b

PR #15, opened 2016-11-29
Issue 439

dannylamb:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/439

This is a first pass at a default rdf mapping for a collection object.  I'm relying pretty heavily on schema.org except for our pcdm predicates.  Let me know if it's too much and we should provide even less.

Here's an example object whose jsonld I've retrieved using this mapping:
```json
{
  "@context": {
    "ldp": "http://www.w3.org/ns/ldp#",
    "dc11": "http://purl.org/dc/elements/1.1/",
    "nfo": "http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/v1.1/",
    "ebucore": "http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#",
    "fedora": "http://fedora.info/definitions/v4/repository#",
    "owl": "http://www.w3.org/2002/07/owl#",
    "ore": "http://www.openarchives.org/ore/terms/",
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "content": "http://purl.org/rss/1.0/modules/content/",
    "dc": "http://purl.org/dc/terms/",
    "foaf": "http://xmlns.com/foaf/0.1/",
    "og": "http://ogp.me/ns#",
    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
    "schema": "http://schema.org/",
    "sioc": "http://rdfs.org/sioc/ns#",
    "sioct": "http://rdfs.org/sioc/types#",
    "skos": "http://www.w3.org/2004/02/skos/core#",
    "xsd": "http://www.w3.org/2001/XMLSchema#"
  },
  "@graph": [
    {
      "@id": "http://localhost:8000/fedora_resource/4?_format=jsonld",
      "@type": [
        "pcdm:Collection",
        "schema:CollectionPage"
      ],
      "nfo:uuid": [
        {
          "@value": "b244d41c-35b7-48c7-92d4-6d1c4f6fa5d1"
        }
      ],
      "schema:name": [
        {
          "@value": "Collection B"
        }
      ],
      "schema:dateCreated": [
        {
          "@value": "2016-11-29T17:52:29+00:00"
        }
      ],
      "schema:dateModified": [
        {
          "@value": "2016-11-29T17:52:29+00:00"
        }
      ],
      "schema:text": [
        {
          "@value": "<p>I'm a child collection</p>\r\n"
        }
      ],
      "pcdm:memberOf": [
        {
          "@id": "http://localhost:8000/fedora_resource/3?_format=jsonld"
        }
      ]
    },
    {
      "@id": "http://localhost:8000/fedora_resource/3?_format=jsonld",
      "@type": [
        "pcdm:Collection",
        "schema:CollectionPage"
      ]
    }
  ]
}
```


whikloj:
I'm wondering if anything here should contain HTML?
```
     "schema:description": [
        {
          "@value": "<p>I'm a child collection</p>\r\n"
        }
      ],
```

dannylamb:
FWIW that's what Drupal is doing with its node content type rdf mappings.  I don't know if that's right or wrong, but it's what they're doing.

Personally, I'm ambivalent.  Is there strong reasoning to make a decision, one way or the other?  I mean, we can change this default to not allow html, but are you saying no one ever can put html formatted content as rdf?  Do we enforce that everywhere?  Or do we let you do what you want, even if there's perceived risk?

whikloj:
@dannylamb I'd say we should try to avoid it, we are talking about the serialized metadata and I think we want to make it clear that when we serialize our content to json-ld (or whatever) that we are not (by default) including display content. If you want to change it up, that is totally cool but not the suggested way.

But perhaps I am overthinking this, I know that I'd prefer not to have that stuff end up in my Fedora properties. But as long as there is a way to remove it, I can do that myself.

dannylamb:
Changing the description field to long plain text will do that, but would lose summary/trimming capabilities.  I don't think it'd be a big deal to implement an extra trimmed field for teasers that's updated automatically.  I'm fine with that being how we roll.

But the question still stands, how far do you go to enforce this?  Using the right fields gets you the right form widget.  I imagine we'd be sanitizing when posting to create an object as well.  But at any point in time someone could still use an html field if they wanted to or by accident.  What do we do in that case?

Documentation?

dannylamb:
https://github.com/Islandora-CLAW/CLAW/issues/441

whikloj:
@dannylamb you raise good points...I'm not sure. Maybe I'm worrying about something in Drupal we should deal with in the microservices?

dannylamb:
@whikloj I think our sensible defaults we provide should be plain text, and we need to make reasonable efforts to stop people from sticking in html in those fields.  But if someone wants a field with html and adds it to an entity, there's not much we can do.  Let's just say I don't think removing that type of field from Drupal to make sure no one ever has markup is worthwhile/sane.

So let's just guide people best we can.

I made Islandora-CLAW/CLAW#441 to get us started.  I imagine soon we'll need to emulate trimmed text functionality once we start displaying lists of teaser content, but we'll cross that bridge when we get there.

DiegoPino:
@whikloj @dannylamb wondering if namespace prefix registering could be done per module. I mean it's ok to go PCDM, but islandora itself does not provide pcdm, collection does. So that hook should be implemented in the collection one to register pcdm. Basically (i know i'm difficult to understand sorry) i want each module to be responsible for it's own namespaces.

whikloj:
I kinda figured we would probably implement some PCDM in islandora and as these other modules would rely on islandora, then we don't have to have each of them re-register the namespace prefix.

dannylamb:
Ah, it's just the prefix.  If it were actual functionality I would totally agree with @DiegoPino.  But let's just put it in.  I mean, we're already registering ebucore and foaf without any assumptions about their use.

dannylamb:
@whikloj Bump.  You cool with that?

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
```diff
@@ -12,24 +12,24 @@ id: fedora_resource.collection
...
     properties:
-      - 'dc:description'
+      - 'schema:text'
   field_memberof:
     properties:
-      - 'dc:isPartOf'
+      - 'pcdm:memberOf'
+  uuid:
+    properties:
+      - 'nfo:uuid'
```
DiegoPino: We already had this talk a year ago... nfo: is for files not rdf resources.

dannylamb: Unless you changed things while I was gone, we've always used nfo:uuid for all resources.

dannylamb: What do you want it to be then?

DiegoPino: That we have used it does not make it right. This comes back to having our own ontology, but in this case in specific it must be semantically correct, whatever you go (even dc identifier would be better) and `nfo` is not. I did changed that,  the whole repository re-organization may not maintained history on that but the proposed alternative was/is there. When i wrote the resource services and PDX part of uuid predicates (I think you were still online to that date) we had this talk. I went for an alternative `isl:` prefix, (also talked about that in multiple claw talks)
https://github.com/search?q=org%3AIslandora-CLAW+isl%3A&ref=searchresults&type=Issues&utf8=✓

Also according to this: "Dani likes isl"
https://github.com/search?q=org%3AIslandora-CLAW+isl%3A&ref=searchresults&type=Wikis&utf8=✓

nfo:uuid is plainly wrong. Domain is nfo:Filesystem

> Type	rdf:Property, rdfs:Resource
Domain	nfo:Filesystem
Range	xsd:string
Superproperties	nie:identifier, dc:identifier, nao:identifier
Subproperties	--
Description	Universally unique identifier of the filesystem. In the future, this property may have its parent changed to a more generic class.

dannylamb: So dc:identifier?  I'm just trying to find out what you want so I can put it in there.

ruebot: @DiegoPino do you have a suggested predicate that @dannylamb can use? I think that is all he is asking for, and would probably be the most productive way to move this review forward.

dannylamb: The other solution is to strike it altogether and let people decide on their own, which I'm also ok with.  This is just defaults, after all.  I have no horse in this race and am cool with what others feel is best.

DiegoPino: @dannylamb @ruebot there is no harm in discussing this or do you need a quick answer?

I thought i was pretty clear why i was citing the "isl" issue with all the links and explanations, took me some effort to do it 😄 just to have this conversation again.

There is code in Microservices that depends on this (even when there are plans to change that or not using them at all, if those plan happen to exists) , so any choice needs some analysis and 
some further updated to code problaby, since it is "the" property we are using to match fedora to drupal and back.

I proposed a #hasURN property and piece of that is part right now of the microservices. It' is waiting for some changes of course and the final ontology. hasURN has some benefits, allows us to define it is a uuid, and if namespaced in islandora, then it's unique to our system (like vector clock thing). 

I like dc:identifier a lot and i would say, lets use it -also-, but it has just a small problem if we go only for that. Domain. A resource can have a lot of local identifiers, and for islandora we need a drupal specific domain-ed one, and just one. If we put the uuid into dc:identifier we are giving our users
* the chance to change uuids easily and breaking that relation
* or the other way, making them adding their own local ones pretty difficult.

If there is no consensus then please ask me to approve the pull as-is and we can discuss this again in the future when the issue appears again. I do take my semantics very serious, sorry about that so would not like to make a suggestion without consensus

dannylamb: I did some digging on the code you're mentioning.  Git blame reveals it was committed while I was out on paternity leave.  Sorry I'm out of touch on that.

Would you prefer we use a hasURN predicate in anticipation of an ontology landing, since OWA will give us problems with dc:identifier?

whikloj: I think @DiegoPino would like to also have something [like this](https://github.com/Islandora-CLAW/pdx/blob/master/src/CollectionService/Controller/CollectionController.php#L77-L82) which we ended up using along with the [nfo:uuid](https://github.com/Islandora-CLAW/pdx/blob/master/src/CollectionService/Controller/CollectionController.php#L71).
I have no problem with this, lets just make sure that the URN we are happy with the namespace before we get tooo far.

dannylamb:
@whikloj @DiegoPino I've updated this PR to use 'islandora:hasURN' in anticipation of the ontology being defined.

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
```diff
@@ -12,24 +12,24 @@ id: fedora_resource.collection
 targetEntityType: fedora_resource
 bundle: collection
 types:
-  - 'schema:Thing'
-  - 'ldp:RDFSource'
-  - 'ldp:Container'
+  - 'pcdm:Collection'
+  - 'schema:CollectionPage'
```
whikloj: We haven't defined the `schema` prefix

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
```diff
@@ -12,24 +12,24 @@ id: fedora_resource.collection
...
 fieldMappings:
   field_description:
     properties:
-      - 'dc:description'
+      - 'schema:text'
```
whikloj: same as above re: `schema`

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
```diff
@@ -12,24 +12,24 @@ id: fedora_resource.collection
...
-      - 'dc:isPartOf'
+      - 'pcdm:memberOf'
+  uuid:
+    properties:
+      - 'islandora:hasURN'
   name:
     properties:
-      - 'dc:title'
-      - 'rdf:label'
+      - 'schema:name'
```
whikloj: same as above re: `schema`

islandora_collection/config/install/rdf.mapping.fedora_resource_type.collection.yml
DiegoPino: @whikloj , RDF module provides schema.org by default. So that should be solved

whikloj: Oh cool, my bad.

DiegoPino:
As discussed in other communication channels, it could be good to keep `dcterms` properties around as default mappings, to help with transition for our 1.x users, because schema.org uses `dcterms` to describe all non `schema.org` specific (http://schema.rdfs.org) and also because some of the PCDM examples in the MVP use `dcterms`.

dannylamb:
@DiegoPino updated PR to include dc that got removed

PR #16, opened 2016-12-02
Add RDF map for fedora:hasParent and ldp:contains

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/442

Check this first, I didn't map every possible field. Which I can if we want. Also there is **no** default mapping for a Non RDF Source. I can add that to this, or we should create another ticket.


dannylamb:
Lets make another ticket for nonrdfresources  @whikloj.

whikloj:
Sounds good

PR #17, opened 2016-12-05
Correctly get parent

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/443

How to test:

First create a Fedora Resource (Content -> Fedora Resource -> Add Fedora Resource) 
**DON'T** set a parent.
Save the new Fedora Resource.

Create a second Fedora Resource and set this parent as the first Fedora Resource you created.
Save the new Fedora Resource.

View the Fedora Resource List (Content -> Fedora Resource), each Fedora Resource lists itself as it's own parent.
ie.
issue-443-1

Checkout this PR, perform cache rebuild (`drupal cache:rebuild all` at drupal root folder), refresh the Fedora Resource List. The parents should be correct.
issue-443-2

### DIFF
--- a/islandora/src/Entity/FedoraResource.php
+++ b/islandora/src/Entity/FedoraResource.php
@@ -267,7 +267,7 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
    */
   public static function getFedoraRoot() {
     // Just stub code, we need to figure out what "root is" in this context.
-    return array('root');
+    return NULL;
   }
 
   /**
@@ -325,7 +325,7 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
       ->setRevisionable(TRUE)
       ->setSetting('target_type', 'fedora_resource')
       ->setSetting('handler', 'default')
-      ->setDefaultValue(NULL)
+      ->setDefaultValue('Drupal\islandora\Entity\FedoraResource::getFedoraRoot')
       ->setTranslatable(TRUE)
       ->setDisplayOptions('view', array(
         'label' => 'hidden',
--- a/islandora/src/Entity/FedoraResource.php
+++ b/islandora/src/Entity/FedoraResource.php
@@ -223,21 +223,30 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
    * {@inheritdoc}
    */
   public function hasParent() {
-    return ($this->get('fedora_has_parent') !== NULL);
+    return ($this->get('fedora_has_parent')->first() !== NULL);
   }
 
   /**
    * {@inheritdoc}
    */
   public function getParent() {
-    return $this->get('fedora_has_parent')->getEntity();
+    return $this->get('fedora_has_parent')
+      ->first()
+      ->get('entity')
+      ->getTarget()
+      ->getValue();
   }
 
   /**
    * {@inheritdoc}
    */
   public function getParentId() {
-    return $this->get('fedora_has_parent')->getEntity()->id();
+    return $this->get('fedora_has_parent')
+      ->first()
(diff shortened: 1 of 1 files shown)


islandora/src/Entity/FedoraResource.php
```diff
@@ -316,7 +325,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setRevisionable(TRUE)
       ->setSetting('target_type', 'fedora_resource')
       ->setSetting('handler', 'default')
-      ->setDefaultValueCallback('Drupal\islandora\Entity\FedoraResource::getFedoraRoot')
```
DiegoPino: Should we "stay" or implement with the idea that `FedoraResource::getFedoraRoot` wanted to implement?
It's  LDP anyway, so there must be at least an  LDP root always

whikloj: Because this field is an entity reference, we need something here and 'root' seemed more of a pain to test for. Especially as we can say that NULL is == to the root of the repository without actually adding a fake entity relationship in.

whikloj: Otherwise I guess we can use the configured root of the fedora resource, but (again I'm new to this) can we have an entity reference field that does not reference an entity or NULL? Do we need to create a node/fedora_resource for "root"?

DiegoPino: @whikloj i agree. That is the drupal way. I just would like to avoid that null. So, as @ruebot said some time in the past, should we go for an install time fedora resource root? (and yes, not only ruebot, you are saying that also). It's just an yaml file at the end, but it's a special one, because you can't create or remove it (sync sync sync) from fedora4

DiegoPino: @whikloj i got an idea. Could be assume that the home page of drupal site is an entity? It's a view most of the time <front> and since in Drupal 8 views are entities (uuid) could that be the default? If it does not sound reasonable or do-able because of time restrictions, i would say then lets move forward and use for now NULL.

whikloj: I'm fine with that. We will have to make sure to correct that once we serialize it to send to Fedora.

dannylamb: I LOVE this idea, but target type is fedora_resource.  We can't make the field reference a view and fedora_resource entities. :(

whikloj: Hrm...okay sooooo what about auto-creating a fedora entity to match the [`fedora_rest_endpoint`](https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/islandora/config/install/islandora.settings.yml#L3) we set on install?

DiegoPino: @whikloj ok, sounds good. But that one needs to be "inmune" to deletion. So fedora_resources could need an "protected" attribute?

dannylamb: @DiegoPino @whikloj Yeah, we're in required_object territory.

And I think there's nothing stopping you from deleting the root collection in 7.x-1.x?  I'm not sure what's the best way to enforce something like that.  Because I agree, you should not be allowed to tamper with it.

whikloj: @DiegoPino @dannylamb Ok, I'm still working the import/export sprint, but I'll do this next.

whikloj:
Ok, so this doesn't sit too long and because I think returning `null` is better than returning `"root"`. I'm going to leave this as is. We can look at whether we want to build in a required object into the system.

DiegoPino:
@whikloj looks good for me. Thanks!!

PR #22, opened 2016-12-08
add fedora resource teaser config schema for issue #447

kimpham54:
Addresses issue [Islandora-CLAW/CLAW#447](https://github.com/Islandora-CLAW/CLAW/issues/447), tested by exporting and importing again, and enabling view mode to Non-RDF source and RDF source.

### DIFF
--- a/islandora/config/install/core.entity_view_mode.fedora_resource.teaser.yml
+++ b/islandora/config/install/core.entity_view_mode.fedora_resource.teaser.yml
@@ -1,0 +1,10 @@
+uuid: 0efc20d3-0bd6-4606-842f-b87548de1924
+langcode: en
+status: true
+dependencies:
+  module:
+    - islandora
+id: fedora_resource.teaser
+label: Teaser
+targetEntityType: fedora_resource
+cache: true


DiegoPino:
@kimpham54 looks good! @Islandora-CLAW/committers should we go further and define Teaser content for bundled entity types or will that be a different ISSUE?
e.g (for RDF Source)
In this case entity type node, bundle page.
https://api.drupal.org/api/drupal/core!profiles!standard!config!install!core.entity_view_display.node.page.teaser.yml/8.2.x

kimpham54:
@DiegoPino @Islandora-CLAW/committers if you want to keep it in the same ticket, I can work on it. Would you like me to take a first stab at styling the teaser view for RDF and non-RDF?

whikloj:
@kimpham54 Currently with this config we get the same teaser view for both Rdf and NonRdf correct?

kimpham54:
@whikloj yep, so right now the Default is the same as the Teaser for both RDF and non RDF source, which is that the name appears as plain text, all other fields are hidden. The Default view mode is also not currently defined.

DiegoPino:
@whikloj yes, because it's entity type based, not on bundle (as in the example i posted which targets a bundled entity type, a.k.a content type)

dannylamb:
@kimpham54 oh, and you get default by default.  It doesn't need to be explicitly defined.  I'm purposefully not basing anything off of that because it's the first thing people play with / break.

dannylamb:
Thx so much @kimpham54

kimpham54:
@dannylamb no problem

PR #26, opened 2016-12-22
Move ldp_contains to rdf_source bundle

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/397

How to test:

1. Uninstall Islandora
1. Apply PR
1. Install Islandora

You should now have a ldp_contains field when creating a Rdf Source Fedora Resource, but not when creating a Non Rdf Source Fedora Resource.


whikloj:
@dannylamb I added the Views wizard, a default entity reference view and changed the ldp_contains field to use this view. 

If this is good it should be easy to make collection also use this same view and be able to reference **all** fedora resource entities.

PR #29, opened 2017-01-11
Revert "Adding @dev where appropriate"

dannylamb:
Reverts Islandora-CLAW/islandora#28

We goofed.  The problem was elsewhere.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,9 +19,9 @@
     }
   ],
   "require": {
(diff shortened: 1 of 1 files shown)


dannylamb:
@whikloj This turns out to have not been neccessary.  Can you take care of this along wtih https://github.com/Islandora-CLAW/islandora_collection/pull/3?

PR #38, opened 2017-02-17
Add JWT Token Authentication

jonathangreen:
This adds JWT token authentication to Islandora. We send these tokens
along with broadcast messages, and can use them to authenticate with
Fedora as well as call back into Islandora.

This is part of:
https://github.com/Islandora-CLAW/CLAW/issues/520

PR #39, opened 2017-02-18
Change how we setup travis to use composer to install all dependencies

jonathangreen:
The way were we installing Drupal for travis before was differnt from
how we installed it in the vagrant box. It required manually adding
dependancies for libraries. This should install all the dependancies
using composer automatically.

Came up on this pull request:
https://github.com/Islandora-CLAW/islandora/pull/38

Figured I should break it out to another pull since it got larger then 
I intended it to be.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,9 @@ php:
   - 5.5
   - 5.6
   - 7.0
-  - hhvm
+  - 7.1
 
 matrix:
-    allow_failures:
-        - php: hhvm
     fast_finish: true
 
 branches:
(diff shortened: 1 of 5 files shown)


jonathangreen:
After this lands, https://github.com/Islandora-CLAW/islandora_collection should probably get the same treatment.

.travis.yml
```diff
@@ -4,32 +4,30 @@ php:
...
+  - cd web; drush en -y islandora
 
 script:
-  - $ISLANDORA_DIR/.scripts/line_endings.sh .
-  - cd $HOME/drupal/modules/islandora
-  - php vendor/bin/phpcs --standard=Drupal --extensions=module/php,php/php src tests
-  - php vendor/bin/phpcpd --names *.module,*.inc,*.test,*.php --exclude=vendor .
-  - cd $HOME/drupal; php core/scripts/run-tests.sh --verbose --php `which php` islandora
+  - $SCRIPT_DIR/line_endings.sh $TRAVIS_BUILD_DIR
+  - phpcs --standard=Drupal $TRAVIS_BUILD_DIR
```
whikloj: I'm not sure we need the old `--extensions=module/php,php/php`, but as the default is 
> ...check any file it finds with a .inc, .php, .js or .css extension, ...

Should we just specify `--extensions=php`?

jonathangreen: good catch. probably not checking enough file types. should call it with the extensions listed here:
https://www.drupal.org/node/1587138

`phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md /file/to/drupal/example_module`

Will update.

jonathangreen: Pushed an update to test all the file types.

jonathangreen:
@DiegoPino I feel like if we want to target a particular version of drupal we should probably do it here: 
https://github.com/Islandora-CLAW/drupal-project/blob/8.x-1.x/composer.json#L23

So it applies to both travis and the vagrant build.

README.md
```diff
@@ -16,10 +16,22 @@ Current maintainers:
...
 
 [GPLv2](http://www.gnu.org/licenses/gpl-2.0.txt)
+
+[1]: https://travis-ci.org/Islandora-CLAW/islandora.png?branch=8.x-1.x
+[2]: http://img.shields.io/badge/CONTRIBUTING-Guidelines-blue.svg
+[3]: https://img.shields.io/badge/license-GPLv2-blue.svg?style=flat-square
+[4]: https://github.com/Islandora-CLAW/CLAW/wiki
+[5]: http://islandora.ca/sites/default/files/islandora_cla.pdf
+[6]: http://islandora.ca/sites/default/files/islandora_ccla.pdf
+[7]: http://islandora.ca/resources/contributors
```
whikloj: Is this a new feature of Github Markdown? 'Cause this is awesome

whikloj:
This is self-testing in that we know islandora passes with the existing Travis setup and you can see how it runs with the new Travis setup. 👍  merging.

README.md
jonathangreen: Seems like its always been a feature. I'd never actually seen it in use before I went looking for a way to make the readme a little more readable though.

DiegoPino:
@jonathangreen are you making pulls for https://github.com/islandora-claw/islandora_collection and https://github.com/islandora-claw/claw-jsonld to level-up/match Travis-CI to this one? Need help on those?

jonathangreen:
Sure. I'll work on it. I made a new issue here so I don't forget:
https://github.com/Islandora-CLAW/CLAW/issues/540

Can someone assign it to me? Thanks.

PR #33, opened 2017-01-30
JSON-LD Context generator

DiegoPino:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/493 
Resolves https://github.com/Islandora-CLAW/CLAW/issues/505 (via caching and cache tags)
update:
**UPDATE:**

Ready to be reviewed and merged

**UPDATE:**

Resolves https://github.com/Islandora-CLAW/CLAW/issues/506 (via controller that responds with application/ld+json)

Base pull here, service, interface, and Class. Missing tests. Some edge cases not covered, probably more complicated than it should. Drupal Coding standards cleaned.
Issue 505 solved via automatic cache invalidation via cache tags.

~~Please, DO NOT MERGE!. I repeat, please DO NOT MERGE!~~ 😝 

~~Update: Ok, now this should do something and could be eventually merged~~

Testing:

if you enable this module and go to
http://localhost:8000/fedora_resource_context/rdf_source
or
http://localhost:8000/fedora_resource_context/rdf_non_source
or
http://localhost:8000/fedora_resource_context/collection
You should see some `@context`being generated

If you feel lucky, you can add more mappings to our YAML bundle mappings and see those changes reflected (remember that config entities are cached, so run always `drupal cache:rebuild`)

Please inform of any issues so I can keep improving this service.

Thanks!

PR #40, opened 2017-02-22
Issue 525 - POST request for FedoraResource entities

Natkeeran:
## What does this Pull Request do?
This PR is a first go at implementing: implements https://github.com/Islandora-CLAW/CLAW/issues/525 

Basically, it is should take a POST request to http://localhost:8000/fedora_resource/ and create a fedora_resource entity bundle.  The type of bundle is specified by the X-Islandora-Bundle header.  And the content or fields of the bundle is specified in the body as expanded json-ld.

## How should this be tested?
* This PR depends on https://github.com/Islandora-CLAW/islandora/pull/33.  While this PR is under review, you can pull that PR.  Then, merge pull this PR!  islandora.services.yml  is the only conflicting file.   
* Launch the POSTMAN
* Create a POST request with the sample body below and X-Islandora-Bundle header value rdf_source.  
```
{
    "@type": [
      "http://www.w3.org/ns/ldp#RDFSource",
      "http://www.w3.org/ns/ldp#Container"
    ],
    "http://purl.org/dc/elements/1.1/title": [
      {
        "@value": "This is a RDF Source"
      }
    ],
    "http://schema.org/dateCreated": [
      {
        "@value": "2017-02-16T20:26:06+00:00"
      }
    ],
    "http://schema.org/dateModified": [
      {
        "@value": "2017-02-16T20:26:06+00:00"
      }
    ]
}
```
* It should create it and return a Json message with the ID of the resource.

## Notes and Questions
* Checksum not implemented. 
* Please review how the rdf -> fields mapping is implemented.  There should be simpler ways!
* If two RDF triples are mapped to same field, which value should we take!
* Can we always assume expanded jsonld?  What are the supported formats?  Is there a library to do the jsonld conversions?
* Should most of the logic move into its own class?


whikloj:
@Natkeeran I'll pull this down and start testing if you want to clean up the code. Looks like Travis is failing on checkstyle stuff.

Natkeeran:
@whikloj will do thank you

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+        $entity = entity_create($entity_type, array('type' => $bundle));
+
+        // Loop through all properties
+        foreach($body as $property => $fieldValue){
+            // This gets auto created!
+            if($property === "@type"){
+                continue;$dump = print_r($variable, true);
+            }
+
+            // Expanded triple $property i.e "http://purl.org/dc/elements/1.1/title": [{ "@value": "This is a RDF Source" }]
```
DiegoPino: Too long comment

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+
+        // If bundle specified in X-Islandora-Bundle is not a valid one, return bad request
+        $bundles =\Drupal::entityManager()->getBundleInfo($entity_type);
+        if (!array_key_exists($bundle, $bundles)) {
+            $response['data'] = "Bundle not found.";
+            return new JsonResponse($response, 400);
+        }
+
+        // Request Body
+        $body = json_decode( $request->getContent(), TRUE );
```
DiegoPino: coding standards

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+        $entity_type = 'fedora_resource';
+
+        // If X-Islandora-Bundle is not set, return bad request
+        if (!isset($bundle) || $bundle == ''){
+            $response['data'] = "X-Islandora-Bundle header not defined";
+            return new JsonResponse($response, 400);
+        }
+
+        // If bundle specified in X-Islandora-Bundle is not a valid one, return bad request
+        $bundles =\Drupal::entityManager()->getBundleInfo($entity_type);
```
DiegoPino: Inject ::entityManager() instead of calling statically via drupal helper

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+        // Loop through all properties
+        foreach($body as $property => $fieldValue){
+            // This gets auto created!
+            if($property === "@type"){
+                continue;$dump = print_r($variable, true);
+            }
+
+            // Expanded triple $property i.e "http://purl.org/dc/elements/1.1/title": [{ "@value": "This is a RDF Source" }]
+            // Gets converted into dc11:title
+            $propertyName = substr($property, strrpos($property, '/') + 1);
```
DiegoPino: This looks a bit overcomplicated and not sure if it covers all use cases of IRI/URI expansion/compaction. Expanding/compacting an IRI/URI should be a method by itself.

src/Controller/FedoraResourcePOSTController.php
DiegoPino: If we are handling JSON-LD as json, are we discarding JSON-LD docs that have`@context` or include `@graph'?

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+            $nameSpaceURL = substr($property, 0, strrpos($property, "/"));
+            $nsPrefix = array_search($nameSpaceURL . "/", $arrNameSpaces);
+            $prefixAndProp = $nsPrefix . ":" . $propertyName;
+
+            $fileName = $this->getFieldName($arrFieldsWithRDFMapping,  $prefixAndProp);
```
DiegoPino: If possible, variable naming should match what they contain. $fileName is tricky here,

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,171 @@
...
+            $value = $fieldValue[0]["@value"];
+            $entity->$fileName = $value;
+        }
+
+        $entity->save();
```
DiegoPino: the save process should take in account the response. Did if really saved? some try catch could help. See ::save () implementation for details

islandora.routing.yml
```diff
@@ -28,3 +28,11 @@ system.islandora_settings:
     _title: 'Islandora Settings'
   requirements:
     _permission: 'administer site configuration'
+
+# POST request for FedoraResource entities
+islandora.post:
+  path: 'fedora_resource'
+  defaults: { _controller: '\Drupal\islandora\Controller\FedoraResourcePOSTController::processPost' }
+  methods:  [POST]
```
DiegoPino: this requires permissions, at least same as creating a resource via UI would required

dannylamb:
@Natkeeran  All of your business logic of creating the entity from the request body should go in https://github.com/Islandora-CLAW/claw-jsonld instead of living in the controller directly.  'Deserialization' is just a `json_decode` like you're doing now, and 'denormalization' is the rest of what you've written.  

Separating that out will make testing easier, too, since you can mock the serializer and its output and just focus on HTTP layer.

Natkeeran:
Thank you.  @DiegoPino @dannylamb Working on addressing your feedback/review.

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -19,16 +21,24 @@ class FedoraResourcePOSTController extends ControllerBase {
   const HEADER_CONTENT_TYPE = 'Content-Type';
   const HEADER_BUNDLE = 'X-Islandora-Bundle';
 
-  public $jsonldGenerator;
+  protected $jsonldGenerator;
```
DiegoPino: Cool. Now you need to add comments for this protected vars also. 😄

src/Controller/FedoraResourcePOSTController.php
Natkeeran: @DiegoPino 
From what I understand, the rdf mapping (RdfMapping::load($entity_type . "." . $bundle)->getFieldMapping($field_name) will only give us compact or prefix:RDFPropertyName of the field.  
  
We can get the expanded version of the rdf mapping of the field name by using the context and this library: https://github.com/digitalbazaar/php-json-ld.  Then, we can compare the expanded IRI/URI of the request with the expanded RIR/URI of the rdf mapping to get the field name.  

Example:

```
    $bundleContext = $this->jsonldGenerator->getContext($entity_type . "." . $bundle);
    $contextInfo = json_decode($bundleContext);

    $doc = (object)array(
      "dc11:title" => "value"
    );

    $compacted = jsonld_compact($doc, (object)$contextInfo);
    $expanded = jsonld_expand($compacted);
```

Will give us:
```
[
    {
        "http://purl.org/dc/elements/1.1/title": [
            {
                "@value": "value"
            }
        ]
    }
]
```

I am trying to avoid writing any expansion/compact algorithm, and reuse an existing library.

Natkeeran:
@DiegoPino @whikloj 

I've made several of the changes requested such as code style cleanup, DI, and additional error checking.  I have not addressed, Diego's concern with respect to Compacting and Expanding URIs.  I am re posting a comment I made earlier below for feedback:
 
From what I understand, the rdf mapping (RdfMapping::load($entity_type . "." . $bundle)->getFieldMapping($field_name) will only give us compact or prefix:RDFPropertyName of the field.  
  
We can get the expanded version of the rdf mapping of the field name by using the context and this library: https://github.com/digitalbazaar/php-json-ld.  Then, we can compare the expanded IRI/URI of the request with the expanded RIR/URI of the rdf mapping to get the field name.  

Example:

```
    $bundleContext = $this->jsonldGenerator->getContext($entity_type . "." . $bundle);
    $contextInfo = json_decode($bundleContext);

    $doc = (object)array(
      "dc11:title" => "value"
    );

    $compacted = jsonld_compact($doc, (object)$contextInfo);
    $expanded = jsonld_expand($compacted);
```

Will give us:
```
[
    {
        "http://purl.org/dc/elements/1.1/title": [
            {
                "@value": "value"
            }
        ]
    }
]
```

Is this a workable approach?  I am trying to avoid writing any expansion/compact algorithm, and reuse an existing library.

DiegoPino:
@Natkeeran looks nice. I'm really a bit tired today so won't be a good code reviewer, but promise to give this a closer look (and some feasible solution) to the expansion issue if needed.  Thanks a lot for your work on this.

dannylamb:
@Natkeeran This work is shaping up nicely, but I _really_ would like to see the bulk of the work go into https://github.com/Islandora-CLAW/claw-jsonld.  Is there anything I can do to help you get started reshaping the work so it functions within the Drupal/Symfony serialization API?

DiegoPino:
@Natkeeran `phpcs` is [complaining a lot](https://travis-ci.org/Islandora-CLAW/islandora/builds/207155449) since the last `drupal standards` update. Since will be working on PHP claw things today, do you need some help with that? Can make a pull against your repo if needed.

DiegoPino:
@Natkeeran related to the expansion/compaction of IRIs and in https://github.com/Islandora-CLAW/islandora/pull/40#issuecomment-283778178 and your request to comment on this on IRC.
I agree. You will just see some implementation notes from here on:

I'm totally in favor of using existing code instead of re-inventing the wheel, moreover, that is the way to do it. 👍 

What we could have available is https://github.com/lanthaler/JsonLD/blob/master/JsonLD.php, which indeed contains the same compaction and expansion logic, and works per document, not per individual IRI. 
It is compliant with the JSON-LD API 1.0 specs, same as digitalbazaar one.

> Side note (THIS IS how expansion, in pseudo code works https://www.w3.org/TR/json-ld-api/#expansion-algorithm)


In concrete, i would suggest a minor refactoring of your code (Ask @dannylamb or @whikloj where that should go, probably different question). Will need @whikloj's ideas here since I divagate too much and I'm also moving back to 7.x-1.x during the release.

1. Make use of the context/prefixes and expand the whole JSON-LD doc you are getting via POST using `/lanthaler/JsonLD`. That is your first filter strategy since you will know for sure if the JSON-LD doc is valid.
2. This is really your own suggestion: you said compare against the expanded version of the field which is right. But you need to expand full documents, not individual URIs when using the JSON-LD libraries which leads to 2 similar options:
   * (OPTIONAL, just to keep in mind) Now you could need some type of framing (fancy word for a JSON-LD document that contains defaults and that you know for sure that works for a given type). This step is optional but would allow you to fill missing default values and validate. Anyway (skip validation if you feel so, out of scope: that is what i am, so my lema). 
   * (NOT optional) Whatever you name this, pseudo-framing (if not doing previous point exactly as described): take all possible RDF-mapped fields for this entity type, add their default values, apply a context/our context and EXPAND, Now you have IRIS to check against your incoming and expanded in point 1. JSON-LD doc.

basically you have 2 docs: both with expanded URI's.

INFINITE. Make entity creation the final step. Only if all previous steps passed.

Hope this helps.

Natkeeran:
Thank you Diego.  I have the required info to move forward.  

Major Tasks
1) Implement Diego's suggestions
2) Refactor to place the code as Danny requested


Additional Notes from IRC for reference:
It is not possible to get expanded to get expanded uri mapping fo the Drupal fields because: 

"Drupal fields do not have URLs attached as properties. So $fields = $this->entityFieldManager->getFieldDefinitions($entity_type, $bundle);) is just a field collection (field class), rdf mapping is in another class and only stores 1. FULL namespace, prefix + URL  multiple prefix:compactedprop per field name"

dannylamb:
@Natkeeran I'd bite this off one chunk at a time.  Let's start by getting the code that makes the entity wrapped up with the serialization api.  You'll be issuing a pull against https://github.com/Islandora-CLAW/claw-jsonld.

It'll go in as a 'denormalizer', which is actually broken down into several smaller bits.  E.g. one class for handling entity references, one for general fields, etc...  Take a look at the hal module for inspiration: https://www.drupal.org/docs/8/core/modules/hal/overview/.  In general, it's like the opposite of what @DiegoPino has already done.

And I'd start by assuming the input to the denormalization is expanded.  We can touch up the controller to handle expanding the input later.  I wouldn't make the serializer responsible for that functionality.

This could easily be a few other tickets, so if you feel like it's scope creeping out of control, I am amenable to breaking this down into smaller pieces.

Natkeeran:
@dannylamb 
I may not be following your terminology (I am referencing https://www.drupal.org/docs/8/api/serialization-api/serialization-api-overview, and https://github.com/Islandora-CLAW/claw-jsonld/tree/8.x-1.x/src).

From your earlier comment, I note: request json -> decode to deserialzie -> array -> denormalize (logic to create entity ) -> object

From @DiegoPino comments, I am planning to implement a class/functions required to get the expanded Json-LD for an entity. That logic needs to go into claw-jsonld.   Is this what you are referring to by serialization?

Where would the actual logic to match and create the entity go?  Into the claw-jsonld as well?

My understanding is that the scope of the ticket is to implement the logic to create an entity from an (expanded) json-ld request for one entity.  The missing piece is what Diego pointed out today.  Of course, we want to do this is in a general way.  Will respond based on your feedback.  Thank you.

dannylamb:
@Natkeeran We're talking past each other a bit here.  We're all pretty much on the same page.

What I'm saying is that the expansion of jsonld probably ought to happen before someone would call `denormalize()`.  That will make the scope of what you have to do in claw-jsonld limited to just creating the entity.

dannylamb:
And after reading that another time, I see I'm totally splitting hairs here.  You are the final arbiter on that decision, so do what feels right.

Natkeeran:
@DiegoPino Have modified the logic to remove the custom expansion of URIs.  Using https://github.com/lanthaler/JsonLD.  Current logic is close to your recommendation, and more robust than before.  

However, you have indicated that there would be default values for the RDF Mapping.  Where can I find that info?  

@dannylamb I will need your help in structuring this.  Once the logic is stable, I will ping you. 

Thank you.

DiegoPino:
@Natkeeran will review as soon as I fix the n+1 issues I'm having with tests on my Pull = which should be, worst scenario considered, tomorrow. Thanks

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,234 @@
...
+    if (!array_key_exists($bundle, $bundles)) {
+      $response['data'] = "Bundle not found.";
+      return new JsonResponse($response, 400);
+    }
+
+    try {
+      // Request Body.
+      $content = json_decode($request->getContent(), TRUE);
+      $newResourceID = $this->createEntity($entity_type, $bundle, $content);
+      $responseData = 'created entity with id ' . $newResourceID;
```
DiegoPino: @Natkeeran @dannylamb not sure if in scope here, but do we could need the UUID after creation back somehow instead/complementary to Integer ID ? in a header maybe? and Also the public URL where this entity will be accessible (imagine a client that is not even aware of the domain of this drupal instance, just hitting some localhost)? Open question, just in case we need that in the future.

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,234 @@
...
+   * @param array $arrFieldsWithRDFMapping
+   *   Field name -> Field mapping array.
+   *
+   * @return array
+   *   Field Mapping -> Default value array.
+   */
+  private function applyDefaultValuesToRdfFields(array $arrFieldsWithRDFMapping) {
+    $arrEntityDocument = array();
+    foreach ($arrFieldsWithRDFMapping as $k => $v) {
+      $arrEntityDocument[$v] = '';
```
DiegoPino: This could need some refactoring, i'm pretty sure that adding an empty string to any RDF-mapped value can bring problems. We can discuss this later in IRC?

tests/fixtures/test.jsonld
```diff
@@ -0,0 +1,22 @@
+{
```
DiegoPino: Where is this fixture used?

Natkeeran:
Thank you Diego.  

Tasks
* apply default values from field definitions
* return what is required based on @dannylamb feedback
* figure our where this needs to go
* (re factor to normalize input request)

src/Controller/FedoraResourcePOSTController.php
dannylamb: @DiegoPino Out of scope for now, but that's a good issue to highlight.  I'll be addressing that as I work through the fedora indexer.

src/Controller/FedoraResourcePOSTController.php
dannylamb: If default values are set in the field config, would that not take over here?  Maybe we should go through the config we have and get that sorted out, then rely on that Drupal behavior.

islandora.routing.yml
dannylamb: This needs the 'add fedora resource entities' permission added to the access requirements.  FYI, if you ever need to look up the string name of permissions, it's in a module's `module_name.permissions.yml` file.  I got this one from https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/islandora.permissions.yml

src/Controller/FedoraResourcePOSTController.php
```diff
@@ -0,0 +1,234 @@
...
+   *
+   * @param string $entity_type
+   *   Entity Type's name.
+   * @param string $bundle
+   *   Bundle's name.
+   *
+   * @return array
+   *   Field to RDF Mapping.
+   */
+  private function getFieldsWithRdfMapping($entity_type, $bundle) {
```
DiegoPino: @Natkeeran. If you are throwing an exception, then you need to use a `@throws` tag in the php doc to make sure people know they have to make some catch/try to avoid their code breaking

Natkeeran:
Outdated, not relevant in the new architecture.

PR #41, opened 2017-02-23
Clean up parents on resource deletion

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/512

Add remove fedora_has_parent function.

Add query to get all inbound references.

Remove fedora_has_parent if the parent is deleted

Edit: Totally forgot

## How to test
Without PR
1. Create a Fedora Resource
2. Create a second Fedora Resource and set the Has Parent field to the first.
3. Delete the parent Fedora Resource
4. Enjoy broken website

With PR
1. Repeat above 1-3
2. Worky-worky

## What changed?

This adds a `removeParent()` function to the `FedoraResourceInterface` as our `setParent()` requires an `EntityTypeInterface` argument and we use `NULL` to remove the value. 

It also adds a overridden `postDelete()` function which performs a query for all `fedora_resource` entities that have a `fedora_has_parent` field with the delete entities ID.
It then removes the child entities (no deleted) parent (using the new `removeParent`)

The same structure could be used to traverse down and remove all children of an entity.

### DIFF
--- a/src/Entity/FedoraResource.php
+++ b/src/Entity/FedoraResource.php
@@ -225,8 +225,7 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
    * {@inheritdoc}
    */
   public function hasParent() {
-    $parent = $this->get('fedora_has_parent')->first();
-    return (!is_null($parent));
+    return (!$this->get('fedora_has_parent')->isEmpty());
   }
 
   /**
@@ -255,7 +254,7 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
   /**
    * {@inheritdoc}
    */
-  public function setParent(EntityTypeInterface $entity) {
+  public function setParent(FedoraResourceInterface $entity) {
     $this->set('fedora_has_parent', $entity);
     return $this;
   }
--- a/src/Entity/FedoraResource.php
+++ b/src/Entity/FedoraResource.php
@@ -291,7 +291,7 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
    *   An array of node ids
    */
   protected static function getInboundReferences(FedoraResourceInterface $entity) {
-    // TODO: Not use static loading
+    // TODO: Not use static loading.
     $query = $query = \Drupal::entityQuery('fedora_resource');
     $query->condition('fedora_has_parent', $entity->id());
     return $query->execute();
--- a/src/Entity/FedoraResource.php
+++ b/src/Entity/FedoraResource.php
@@ -120,6 +120,28 @@ class FedoraResource extends ContentEntityBase implements FedoraResourceInterfac
   /**
    * {@inheritdoc}
    */
+  public static function postDelete(EntityStorageInterface $storage, array $entities) {
+    parent::postDelete($storage, $entities);
+
+    // This removes references to deleted parents. But this could be
+    // used to remove child nodes instead.
+    foreach ($entities as $entity) {
+      if ($entity->getEntityType()->getBundleEntityType() == 'fedora_resource_type') {
+        $references = self::getInboundReferences($entity);
+        if ($references) {
+          $nodes = $storage->loadMultiple($references);
+          foreach ($nodes as $node) {
+            $node->removeParent();
+            $node->save();
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function getType() {
     return $this->bundle();
   }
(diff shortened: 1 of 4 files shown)


src/Entity/FedoraResource.php
```diff
@@ -290,6 +321,22 @@ public function getVclock() {
...
   /**
+   * Get all the objects that have $entity as a fedora_has_parent.
+   *
+   * @param \Drupal\islandora\FedoraResourceInterface $entity
+   *   The entity to find inbound references for.
+   *
+   * @return array
+   *   An array of entity ids
+   */
+  protected static function getInboundReferences(FedoraResourceInterface $entity) {
```
DiegoPino: Love this part, @whikloj++

src/Entity/FedoraResource.php
```diff
@@ -270,6 +293,14 @@ public function setParent(EntityTypeInterface $entity) {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function removeParent() {
```
DiegoPino: Not sure, we probably need to check if some caching would need to be cleared here:
Other ::postDelete implementations do that a lot.

src/Entity/FedoraResource.php
```diff
@@ -270,6 +293,14 @@ public function setParent(EntityTypeInterface $entity) {
...
+    $this->set('fedora_has_parent', NULL);
```
dannylamb: You need to save() right here or this won't take effect.

whikloj:
I'm not sure if @DiegoPino is right about caching, but in my tests I found I had to load the child entity to see changes appear on it when the parent was deleted, I couldn't continue to use the originally created variable.

For example, if you stick the line
```
$this->assertTrue($this->childEntity->get('fedora_has_parent')->isEmpty(), "Child should not have a parent.");
```
[here](https://github.com/Islandora-CLAW/islandora/pull/41/files#diff-1e52c9f468747605e718b058e9a46e14R95) it will fail. 

Even those both `$new_child` and `$this->childEntity` should be referencing the same thing.

Perhaps caching, or maybe something else. I couldn't figure it out. Maybe you all can?

whikloj:
Ok, I'm frustrated. I have phpcs version 2.8.1 and drupal/coder 8.2.11 and it does not show this error. What am I doing wrong??

whikloj:
I should clarify, I do see the error. But for some reason phpcs is not doing its job on my mac.

DiegoPino:
@whikloj did you link to the new standards after installing code 8.2.11? `phpcs --config-set installed_paths /path/to/your/coding/standards?`
I always forget that so just in it  case that happened to you (better details on) https://packagist.org/packages/drupal/coder

Also, those new coding standards have some issues with indenting. The are requiring some extra spaces i never saw before that make code even more complicated to follow. I hope next release fixes the many issues or, i/we get used to that.. whatever is easier.

Lastly, sometimes its just too much to have to learn all this rules for just one of the many programming languages we need to manage on our daily work! 😁  like 
https://github.com/klausi/coder/commit/8f96f4a18901c8d384a6b96202150b866ea77214#diff-acd213be043ff7daab22adb6e6d75784R59

**Did you know you can use longer than 80 chars lines is using `@see` ?** 🏋️

whikloj:
I did set that config path and this was an indenting issues on function doc comments, but I don't understand how one phpcs finds it and the other doesn't.

DiegoPino:
@whikloj maybe packagist, drupal and git versions are not in sync?

whikloj:
@DiegoPino maybe, I un-installed the homebrew version and just installed straight from composer in order to match up. But maybe it has more to do with OSX than the phpcs version.

dannylamb:
@whikloj Did you ever figure out what the deal was with phpcs?  Is this ready for review?  Because it looks good, and if I can put it through the paces I'll merge it.

Thx so much for adding tests.
<image src="https://m.popkey.co/e82fbc/GeQo9.gif"/>

whikloj:
@dannylamb I didn't find out, also I'm seriously wondering if my issue in the test mentioned [here](https://github.com/Islandora-CLAW/islandora/pull/41#issuecomment-285837547) might be due to the need for some cache clearing on the referencing objects. 

Soooo I'm not sure, I'm dedicating Fridays to community work. So if this can sit, I'll see if some cache clearing fixes that issue. Otherwise, merge away and I'll have a look after the fact and if there is a caching change need include it as a separate PR.

dannylamb:
Yeah, I saw you reloading entities in the tests.  If you think something is fishy, I'd prefer to wait until you sort it out.

whikloj:
@dannylamb @DiegoPino ready

PR #43, opened 2017-02-28
Fix issue with travis running force pushed commits

jonathangreen:
TravisCi provides the $TRAVIS_COMMIT environment variable which I was
using to get composer to install the correct version. However it appears
that under some circumstances travis does not set this variable to the
correct commit hash, which can cause our builds to fail. This avoids
using that variable all together to avoid that problem.

Related to: 
https://github.com/Islandora-CLAW/islandora/pull/38
https://github.com/Islandora-CLAW/CLAW/issues/540

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,9 @@ before_install:
 
 install:
   - $SCRIPT_DIR/travis_setup.sh
+  - git -C "$TRAVIS_BUILD_DIR" checkout -b travis-testing
   - cd $DRUPAL_DIR; composer config repositories.local path "$TRAVIS_BUILD_DIR"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,8 +19,9 @@ before_install:
 
 install:
   - $SCRIPT_DIR/travis_setup.sh
+  - git -C "$TRAVIS_BUILD_DIR" checkout -b travis-testing
```
DiegoPino: this line is so clever. 👏

DiegoPino:
@jonathangreen will merge tomorrow together with the other 2 ones ( early in the morning) to give other people time to look at it tonight. Thanks

dannylamb:
@DiegoPino I'm fine with merging now.  Would you like to do the honors?

PR #46, opened 2017-03-09
Issue 526

dannylamb:
**GitHub Issue**: Part of Islandora-CLAW/CLAW#526

* Other Relevant Links:
  * https://github.com/Islandora-CLAW/islandora_collection/pull/14
  * Islandora-CLAW/CLAW#549

Pulls out the tracking of entity updates into its own service.

# What's new?
The number of times an entity has been updated is no longer stored as a field.  A service acts upon the entity during the core entity CRUD hooks. Because it's not a field, this removes it from having to be exposed in the ontology.  

I also rename vclock to version counter all over the place.

* Does this change require documentation to be updated? No
* Does this change add any new dependencies? No
 (ie. Regeneration activity, etc.)? You'll need to uninstall and re-install the module, since there's a hook_schema implementation in the .install file.

# How should this be tested?

The kernel tests should take care of it, but you can manually run through the same steps.  By running `$ drupal database:client`, you can obtain a connection to the drupal database.  You can use the query `SELECT * from islandora_version_count;` to inspect the version count table between each step.
* Create an entity through the UI (this will initialize its row in the db to 0)
* Update the entity through the UI (this will increment its row in the db to 1)
* Delete the entity through the UI (this will delete its row in the db)

Also, if you request jsonld for a RdfSource via REST, you shouldn't see an entry for `islandora:vclock`.
 
# Interested parties
@Islandora-CLAW/committers


DiegoPino:
@dannylamb huge pull! Will review during today and maybe a bit tomorrow also. Looks good.

dannylamb:
Coding standards are really muddying things up there.  Phpcs got super picky recently.

DiegoPino:
@dannylamb someone updated the drupal standards file that phpcs uses. That someone is to blaim 👍

whikloj:
General question here, I noticed you are replacing instances of `array()` with `[]`. Is this conscious and if so, should we make it a "rule" to use the short form for consistency sake?

DiegoPino:
@whikloj they introduced short term enforcement now as part of the Drupal version check https://www.drupal.org/node/2857830 you we are kinda obligated now on 8.x, not on 7.x

islandora.module
```diff
@@ -82,5 +84,41 @@ function islandora_rdf_namespaces() {
...
     'islandora' => 'http://islandora.ca/CLAW/',
     'pcdm' => 'http://pcdm.org/models#',
-  );
+  ];
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function islandora_fedora_resource_insert(EntityInterface $entity) {
```
whikloj: You did want to go the event listener route? I guess this is specific to the type of entity, which I can't see in the event types. Just wondering if there are any tradeoffs between these two options?

dannylamb: I do, but you can't. Responding to core entity CRUD operations is still done with hooks, unfortunately. That or within the entity class itself, like we were doing. I pulled it out into a hook so we can apply version counters to entities we don't have control over, like Media.

whikloj: This could be my confusion between entities and entity types, but can you use these? [`EntityTypeEvents::CREATE`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityTypeEvents.php/constant/EntityTypeEvents%3A%3ACREATE/8.2.x), [`EntityTypeEvents::UPDATE`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityTypeEvents.php/constant/EntityTypeEvents%3A%3AUPDATE/8.2.x) and [`EntityTypeEvents::DELETE`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityTypeEvents.php/constant/EntityTypeEvents%3A%3ADELETE/8.2.x)

DiegoPino:
@dannylamb i have questions about the use of those [hooks](https://github.com/Islandora-CLAW/islandora/pull/46#discussion-diff-105540022R91). First, I do know there are no core events for individual entity crud operations, so i get the problem.

First some background: this [link](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Entity!entity.api.php/group/entity_crud/8.2.x) explains all the hooks and steps called in each CRUD operation and lists also available methods.

So my thoughts are:
1. Right now you are doing `_update`, `_insert` and `_delete` invokations  using `hook_ENTITY_TYPE_` alternative  for a single entity type which is our `fedora_resource` and controlled by us. For this case, seems like that is better done directly in the crud ::methods inside the entity type definition class because we control the entity and we could eventually even dispatch our own events from there.

2. If the justification for using hook_ENTITY_TYPE_* is that we want to add versions to other entity types (reading your other issue about media entities(, then my questions is which ones other than media entities and does that mean that we are going to have to code all possibles here in this module in the future?
Seems to be more reasonable to go for `hook_entity_insert()`,` _delete`, `_update` and inside those functions check which types have islandora versioning enabled (see at the end of this comment the moderation module example)

3. If this is needed for the MVP in the way it is handled here, we are opening a complicated-to-close door, this will inevitable lead to: People will want to make any other content entity islandora-versioning aware in the future and they will have to copy/paste code to do so. (3x invokations per entity type?)
Also, i would even guess the current idea (reimplement for each entity type the same versioning logic, in this case at least fedora_resource and media entities) will get us in trouble with duplicated code detector (copy pasta detector) when running Travis.

**This is the option i feel is safer, still with hooks.**

https://api.drupal.org/api/drupal/core%21modules%21content_moderation%21content_moderation.module/8.2.x (see the source code please, its pretty simple)
 And the tests so see how this works
https://api.drupal.org/api/drupal/core%21modules%21content_moderation%21tests%21src%21Kernel%21EntityOperationsTest.php/8.2.x

Content moderation module seems to be doing the same we are (a set of states associated to existing entities), but in a more general way 

that (optionally) mixed with this maybe:
```Php
Drupal\Core\Config\Entity\ThirdPartySettingsInterface;

/**
 * Interface for configuration entities to store third party information.
 *
 * A third party is a module that needs to store tightly coupled information to
 * the configuration entity. For example, a module alters the node type form
 * can use this to store its configuration so that it will be deployed with the
 * node type.
 */
````

Like here https://api.drupal.org/api/drupal/core%21modules%21content_moderation%21src%21EntityOperations.php/function/EntityOperations%3A%3AupdateOrCreateFromEntity/8.2.x

Could lead to a "no need to code" way of incorporating entities to your versioning schema.

I added this outside the review thread because its not tied to a particular piece of code nor requires a change, just a general comment on how this affects architecture. Mostly open discussion and also to hear what you all think.

islandora.module
dannylamb: Sadly, no.  When first googling I ran into those too, and I was filled with a false sense of hope.  But they're for CRUD operations on entity types, not entities. :(

dannylamb:
@DiegoPino Just trying to put all the stuff that updates version counters in one place since we'll be doing Media entities as well.  Was opting for bundle specific hooks because I think it makes a nice clean the separation.  And don't worry, if phpcpd gets triggered when adding Media, I'll refactor the flagged code into a utility function and call that instead.

You also need to know that this code is for conditional updates.  And although sync will use conditional updates, this is not part of some defining trait (or state as you're reasonably suggesting) that will dictate that process.  It's just so that when PATCHing I can say "Have you changed since last I checked?  Otherwise I'll try again later to make sure I have the newest version."

whikloj:
Tested, works as described. 👍  

@DiegoPino once your concerns are addressed you can merge.

PR #48, opened 2017-03-20
PHP has a odd notion of truth

dannylamb:
**GitHub Issue**: Islandora-CLAW/CLAW#556

Updates the conditional to check if there's a JWT token to adhere to PHP's ridiculous truthiness.

# What's new?
I changed a direct `NULL` comparison to `empty()`.  This way if JWT configuration is invalid, you'll at least get a flash message and a log entry.

# How should this be tested?

Delete your Key or Secret configuration and then perform a CRUD operation on a FedoraResource entity.  You should see an error message and a watchdog log entry.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/src/Plugin/RulesAction/Broadcaster.php
+++ b/src/Plugin/RulesAction/Broadcaster.php
@@ -126,10 +126,10 @@ class Broadcaster extends RulesActionBase implements ContainerFactoryPluginInter
       drupal_set_message(
         t('Error getting JWT token for message. Check JWT Configuration.'), 'error'
       );
+      return;
     }
-    else {
-      $headers['Authorization'] = "Bearer $token";
-    }
+
+    $headers['Authorization'] = "Bearer $token";
 
     // Transform message from string into a proper message object.
     $message = new Message($message, $headers);
--- a/src/Plugin/RulesAction/Broadcaster.php
+++ b/src/Plugin/RulesAction/Broadcaster.php
@@ -118,17 +118,17 @@ class Broadcaster extends RulesActionBase implements ContainerFactoryPluginInter
 
     // Include a token for later authentication in the message.
     $token = $this->auth->generateToken();
-    if ($token !== NULL) {
-      $headers['Authorization'] = "Bearer $token";
-    }
-    else {
+    if (empty($token)) {
       // JWT isn't properly configured. Log and notify user.
       \Drupal::logger('islandora')->error(
-        'Error getting JWT token for message: @msg', ['@msg' => $e->getMessage()]
+        'Error getting JWT token for message: @msg', ['@msg' => $message]
       );
       drupal_set_message(
         t('Error getting JWT token for message. Check JWT Configuration.'), 'error'
       );
+    }
+    else {
+      $headers['Authorization'] = "Bearer $token";
     }
 
     // Transform message from string into a proper message object.
--- a/tests/src/Kernel/BroadcasterTest.php
+++ b/tests/src/Kernel/BroadcasterTest.php
@@ -89,7 +89,7 @@ class BroadcasterTest extends IslandoraKernelTestBase {
         "IslandoraBroadcastRecipients header must be a comma separated list of endpoints"
       );
       $this->assertTrue(
-        strcmp($headers['Authorization'], 'some_token') == 0,
+        strcmp($headers['Authorization'], 'Bearer some_token') == 0,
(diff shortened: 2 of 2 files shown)


dannylamb:
Maybe we want to return here, too?  That would prevent the message from ever getting published to the queue.

jonathangreen:
I think I like the idea of just returning @dannylamb, its not going to do much good sending out a message with no token anyway... 

Kind of brings up the question of how we will do a "reindex" when things fail, as you could create a number of objects, and later need to have them indexed, if things have failed, but thats outside the scope of this.

dannylamb:
@jonathangreen  Yeah, let's just return.  Checking watchdog, i'm spamming the api 10 times to try and get a resource when un-authenticated.  Probably best to avoid that altogether.

For re-indexing, I was looking into using Views Bulk Operations to do that sort of thing, but haven't done enough exploration to speak strongly to the validity of that approach.  Looks like it might be butting heads with the Batch API at the moment: https://www.drupal.org/node/1823572

jonathangreen:
@dannylamb if you want to add the return I'll test and merge this.

src/Plugin/RulesAction/Broadcaster.php
```diff
@@ -118,18 +118,18 @@ protected function doExecute($message, array $recipients) {
 
     // Include a token for later authentication in the message.
     $token = $this->auth->generateToken();
-    if ($token !== NULL) {
-      $headers['Authorization'] = "Bearer $token";
-    }
-    else {
+    if (empty($token)) {
```
whikloj: So I haven't tested this yet, but according to the JWT code, `generateToken()` returns a token or a `FALSE`. Should this be `if ($token === FALSE) {` ??

For reference, the JWT generateToken() function
http://cgit.drupalcode.org/jwt/tree/src/Authentication/Provider/JwtAuth.php#n110
which calls the transcoder->encode() function
http://cgit.drupalcode.org/jwt/tree/src/Transcoder/JwtTranscoder.php#n189

jonathangreen: I agree it should return FALSE. I wonder if the looser test for empty is better though. 

A blank string isn't a valid JWT. Even if for some crazy reason we tried to use the signature type none, it should still be base64(headers).base64(body).(empty string)

So we might as well pop up an error here if the JWT is missing somehow.

dannylamb: I was using empty to catch a blank string, but if FALSE is in the docs, FALSE it shall be.

dannylamb: Ha, I should clear my cache before posting.  Didn't see @jonathangreen's comment.  How about I leave loose with empty and add the return?

whikloj: sure, if everyone else is happy.

dannylamb:
Oh bother... that return statement breaks tests...  I'll investigate.

dannylamb:
So we're failing tests because we're not publishing due to the return, which is getting executed now that the conditional has changed.  Which we do because the config isn't set up right for JWT, which is why I issued this PR in the first place, LOL.

<image src="http://www.reactiongifs.com/wp-content/uploads/2013/10/tim-and-eric-mind-blown.gif" />

So I'll have to figure this out for real and then patch up Travis to include JWT config as well.  Gonna take a while.  Guess this is why we have tests, though.

dannylamb:
Ok, and I've finally unloaded all of my stuff on my current vagrant w/r/t Alpaca and Hypercube.  Time to cycle vagrant to double check my build.  I'll let you know how it goes @whikloj

dannylamb:
Circling back to this: TL;DR we need to add JWT to Travis before tests will pass, since what's blowing up is an actual integration test.  That, or we sneakily remove the test, bwahahahaha....

jonathangreen:
Instead of adding JWT to travis, should we have our test inject the necessary configuration, so that we don't hit this return statement? That way we don't rely on a certain configuration to run our tests.

dannylamb:
@jonathangreen Yeah, that makes sense.  I think we'll need to create a token that can always be validated.  Might take me a bit to figure that out.

@ruebot if you wanna take a stab at it, feel free to issue a pull on pull.

dannylamb:
Ah no... I'm out to lunch.  I'll just mock that thing.

PR #59, opened 2017-05-08
Config for https://github.com/Islandora-CLAW/CLAW/issues/618

dannylamb:
**GitHub Issue**: Part of https://github.com/Islandora-CLAW/CLAW/issues/618

https://github.com/Islandora-CLAW/islandora_collection/pull/20 and https://github.com/Islandora-CLAW/islandora_image/pull/13 must be tested together with this.

Commits exported config used for a larger refactor to make islandora_image a bit more functional in preparation for Islandoracon.  It also makes hasParent hidden on forms and display until we can sort out specifics on its behaviour without confusing folks who are demoing it.

# What's new?
LOTS OF YAML.  And one tweak to a base field definition.

# How should this be tested?

See https://github.com/Islandora-CLAW/CLAW/issues/618 for testing instructions.

# Interested parties
@Islandora-CLAW/committers


dannylamb:
Nooooooooooo..... phpcs why u fail now?

DiegoPino:
@dannylamb I can be mistaken, but are we using phpcs 3.x? 
That is a crazy  version

Seems like that is the default now `composer require squizlabs/PHP_CodeSniffer
Using version ^3.0 for squizlabs/PHP_CodeSniffer
` and  confirming yes 3.0.0 is now official

That version has some  autoloading issues since January 2017 and probably not playing well. What about fixing it to 2.9x ?
composer require squizlabs/PHP_CodeSniffer 2.9

dannylamb:
@DiegoPino https://github.com/Islandora-CLAW/CLAW/pull/619

PR #63, opened 2017-05-31
Issue 643

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/643

Removes our custom FedoraResource entity and replaces our configuration yaml to use nodes instead.

# What's new?
I torched an insane amount of code and config.

# How should this be tested?

See https://github.com/Islandora-CLAW/CLAW/issues/643#issuecomment-305206639 for testing instructions.

# Interested parties
@Islandora-CLAW/committers


whikloj:
@dannylamb looks like the RestResourceConfig doesn't work in Travis?

whikloj:
Ohhhh that is because our Travis setup needs to change to support Nodes now :man_facepalming:

dannylamb:
@whikloj It may be a timing issue during hook_install.  Easy fix is for me to enable `rest` first and then enable the other modules to guarantee it's around.

PR #64, opened 2017-06-01
Using hook_modules_installed instead of hook_install to force rest co…

dannylamb:
…nfig

**GitHub Issue**: Islandora-CLAW/CLAW#643

Moves forcing `rest` config for nodes from `hook_install` to `hook_modules_installed`.

# How should this be tested?

A happy Travis is all we need here.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/islandora.module
+++ b/islandora.module
@@ -41,7 +41,13 @@ function islandora_help($route_name, RouteMatchInterface $route_match) {
 function islandora_modules_installed($modules) {
   // Ensure the auth and serialization formats we need are available.
   if (in_array('rest', $modules)) {
-    $rest_config = RestResourceConfig::load('entity.node');
+    $rest_resource_config_storage = \Drupal::service('entity_type.manager')->getStorage('rest_resource_config');
+
+    $rest_resource_config = $rest_resource_config_storage->load('entity.node');
+
+    if (!$rest_resource_config) {
+      $rest_resource_config = $rest_resource_config_storage->create(['id' => 'entity.node']);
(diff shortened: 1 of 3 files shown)


islandora.module
```diff
@@ -27,14 +28,33 @@ function islandora_help($route_name, RouteMatchInterface $route_match) {
...
     default:
   }
 }
 
 /**
+ * Implements hook_modules_installed().
+ */
+function islandora_modules_installed($modules) {
+  // Ensure the auth and serialization formats we need are available.
+  if (in_array('rest', $modules) {
```
whikloj: missing `)`

dannylamb:
@whikloj derp

dannylamb:
Just realizing now that there's two cases we need to handle: when `rest` is already enabled and when `rest` is enabled along with islandora.

In its current state, Travis fulfills the "rest already enabled" case, so let's make Travis green with that.  Then we'll revert https://github.com/Islandora-CLAW/CLAW/pull/652 and make sure it's green there too.

_Then_ we can resolve Islandora-CLAW/CLAW#643.

dannylamb:
@ruebot You mind merging the revert pull so we can test what happens if you _don't_ have `rest` enabled beforehand?

dannylamb:
@ruebot oh, you did already.  Thanks.

dannylamb:
@ruebot @whikloj We're green on before and after the revert.  This should be good to go now.

PR #74, opened 2018-01-29
Media source update

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/787

Provides an endpoint for you to PUT binaries and update files associated with Media.

# What's new?

* New controller
* New service for business logic required by the controller
* Updated routing and services yml files
* Functional tests

# How should this be tested?

Create a media and give it a file.  Let's say it's created at `/media/1`.  Then PUT a new file to `/media/1/source` like so: `curl -u admin:islandora -v -X PUT -H 'Content-Type: image/png' -H 'Content-Disposition: attachment; filename="my_image.png"' --data-binary @my_image.png localhost:8000/media/1/source`

* The source file should get updated
* Any fields extracted from the file should get updated as well (mimetype, image width/height, etc...)

Some things to note:
- Requests without a `Content-Type` header are rejected with 400.
- Requests without a `Content-Disposition` header are rejected with 400.
- Requests without a file name in the `Content-Disposition` header are rejected with 400.
- Requests without a body are rejected with 400.
- Requests from users without the 'update media' permission are rejected with 403.

# Interested parties
@Islandora-CLAW/committers @mjordan @rosiel


Natkeeran:
Tested this.  It replaced the file with the provided file.  When you go to the media url, it displays the new updated file.  However, the TN displayed in the drupal is still the old one.  

Also, the request without Content-Type still seem to get through.  I did not get the 400 error.

dannylamb:
Looks like cURL automatically generates a `Content-Type` header of `application/x-www-form-urlencoded` if none is provided.

```
daniel@daniel-Latitude-3560:~/Pictures$ curl -u admin:islandora -v -X PUT -H "Content-Disposition: attachment; filename=\"test.jpg\"" --data-binary @test.jpg localhost:8000/media/12/source
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
* Server auth using Basic with user 'admin'
> PUT /media/12/source HTTP/1.1
> Host: localhost:8000
> Authorization: Basic YWRtaW46aXNsYW5kb3Jh
> User-Agent: curl/7.47.0
> Accept: */*
> Content-Disposition: attachment; filename="test.jpg"
> Content-Length: 175298
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
* We are completely uploaded and fine
< HTTP/1.1 204 No Content
< Date: Wed, 31 Jan 2018 15:56:08 GMT
< Server: Apache/2.4.18 (Ubuntu)
< X-Powered-By: PHP/7.0.22-0ubuntu0.16.04.1
< Cache-Control: must-revalidate, no-cache, private
< X-UA-Compatible: IE=edge
< Content-language: en
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Expires: Sun, 19 Nov 1978 05:00:00 GMT
< X-Generator: Drupal 8 (https://www.drupal.org)
< Content-Type: text/html; charset=UTF-8
< 
* Connection #0 to host localhost left intact
```

Do we want to reject with 400 for that mimetype as well?  Sort of makes sense because that mimetype is for form contents from POSTs.

Natkeeran:
With the last commit, works as expected.

mjordan:
@dannylamb if curl does it, maybe we should be checking other common HTTP clients like Guzzle and requests.

dannylamb:
@mjordan  We discussed this at our last call and felt it would be best not to hard-code reject any particular content type, but in the future we can let the user configure a global black list for more control.

I touched up the tests to make them more accurate.  I was sloppily sending requests without a lot of headers instead of ensuring they had all the required headers _except_ the one we're testing for.  Looks like Guzzle doesn't add anything.

PR #77, opened 2018-02-16
Use Context to allow altering the normalized JSON-LD

whikloj:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/768

~Depends on~ https://github.com/Islandora-CLAW/jsonld/pull/24
Actually this just uses the hook from jsonld, not sure if anything bad happens if this is merged first.

Creates a reaction to alter the JSON-LD and add the entity id using a configurable predicate.

# What's new?

NormalizerAlterReaction is an abstract class to base reactions that alter the jsonld on.

MappingUriPredicateReaction extends the new above abstract class.

We call the new `hook_jsonld_alter_normalized_array` from jsonld and call any reactions that are a subclass of the `NormalizerAlterReaction` and they can edit the jsonld normalized array before it is encoded and returned.

* Does this change require documentation to be updated? no
* Does this change add any new dependencies? no
 (ie. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? no

# How should this be tested?

Functional test is included.

But if you pull in the jsonld PR above and this one. You can add a reaction `map_uri_predicate` to an existing or new context, set your predicate and when you get a resource as jsonld it will add that predicate with the URI of the resource.

For instance if I configure the predicate as `owl:booyah` then when I grab `http://localhost:8000/node/1?_format=jsonld` I see this in the json
```
      "http://www.w3.org/2002/07/owl#booyah": [
        {
          "@value": "http://localhost:8000/node/1?_format=jsonld"
        }
      ]
```

# Interested parties
@dannylamb @mjordan @Natkeeran @DiegoPino

### DIFF
--- a/islandora.module
+++ b/islandora.module
@@ -181,3 +181,13 @@ function islandora_entity_delete(EntityInterface $entity) {
     }
   }
 }
+
+/**
+ * Implements hook_jsonld_alter_normalized_array().
+ */
+function islandora_jsonld_alter_normalized_array(EntityInterface $entity, array &$normalized, array $context) {
+  $context_manager = \Drupal::service('context.manager');
+  foreach ($context_manager->getActiveReactions('\Drupal\islandora\ContextReaction\NormalizerAlterReaction') as $reaction) {
+    $reaction->execute($entity, $normalized, $context);
+  }
+}
--- a/src/ContextReaction/NormalizerAlterReaction.php
+++ b/src/ContextReaction/NormalizerAlterReaction.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Drupal\islandora\ContextReaction;
+
+use Drupal\context\ContextReactionPluginBase;
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Base class to alter the normalizes Json-ld.
+ *
+ * Plugins must extend this class to be considered for execution.
+ *
+ * @package Drupal\islandora\ContextReaction
+ */
+abstract class NormalizerAlterReaction extends ContextReactionPluginBase {
+
+  /**
+   * This reaction takes can alter the array of json-ld built from the entity.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface|null $entity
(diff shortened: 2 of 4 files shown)


whikloj:
Ugh test passes locally but fails on Travis, I'll have to fork and start debugging this could take some time.

whikloj:
No I'm wrong, it fails because there is no hook in jsonld yet to do the alteration I'm testing for. So once https://github.com/Islandora-CLAW/jsonld/pull/24 is merged then this should pass. So clearly that should be merged first.

dannylamb:
I just made a context for "Everything" and utilized the reaction successfully :+1:
screenshot from 2018-03-08 10-42-29

After making a collection, I requested it with jsonld and got schema:url (which is what I was expecting):
screenshot from 2018-03-08 10-41-59

dannylamb:
Re-running Travis after merging https://github.com/Islandora-CLAW/jsonld/pull/24

dannylamb:
@whikloj Travis is not appeased :sob: Looks like you're getting 400 instead of 200 at the end of your test.

whikloj:
@dannylamb I'm digging into this, some sort of Guzzle hates me error occurring.

whikloj:
Oooooo I namespaced my reaction, did I update the `reaction_id` in my test? 

Nooooooooooooo

whikloj:
Related, tests for Islandora take 30 minutes to run. We have **GOT** to speed that up.

whikloj:
@dannylamb ready for merge

dannylamb:
@whikloj Yeah, so that's since we're doing functional tests.  There's maybe a way to run in parallel?

For the time being, instead of writing a lot of little test functions, I write one big honkin' one so we only bootstrap Drupal once. It doesn't always make sense to do that, but I try to when it's feasible to drop execution time.

whikloj:
@dannylamb It seems like all discussion around parallelism is in Unit tests. The bootstrapping of multiple Drupal sites to do functional tests seems problematic. We may just have to suffer through for now.

PR #78, opened 2018-02-16
provide namespace configuration

Natkeeran:
**GitHub Issue**: (link)

* Other Relevant Links
```
14:27 <Natkeeran> whikloj: is there any way to add new prefixes
14:27 <Natkeeran> whikloj: example dbpedia-owl, custom
14:28 -!- dwilcox [~dwilcox@134.41.58.226] has quit [Quit: My MacBook Pro has gone to sleep. ZZZzzz…]
14:28 <whikloj> Natkeeran: yes through a hook, we could probably refactor it to use an admin page if you want to build one :)
14:28 <whikloj> https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/islandora.module#L68
14:29 <Natkeeran> whikloj: Is this dynamic, meaning, jsonld pulls it when it needs it?
14:29 <Natkeeran> whikloj: yes, I can make that configurable
14:30 <whikloj> Natkeeran: that is used by the rdf module and in the serializer we make use of the rdf_get_namespaces function to get an array of prefix => uri
14:31 <whikloj> Natkeeran: I don't know if rdf_get_namespaces calls all hooks each time, I imagine there is some cache we would need to clear
14:31 <Natkeeran> whikloj: basically, If I register an array variable and set that array variable via admin page, after cache clear, that should do it, right?
```

Currently, if user uses non configured prefixes in the rdf mapping, resulting jsonld does not contain the the expanded uri for the properties.  This can complicate quering.  The prefix -> namespace mapping is currently hard coded in here:  https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/islandora.module#L68

This PR attempts to make that array configurable via UI.  

# What's new?
This PR changes the hard coded array in `islandora_rdf_namespaces` to a dynamically created array from a string configured via UI.

* Need to verify if this the best approach to represent array.
* Need to verify that performance implications are acceptable.  

# How should this be tested?
* [Uninstall islandora](https://gist.github.com/whikloj/614fa21d7145b09197fd1e8513dd2f41)
* Install this PR
* Verify that default namespaces are set here: http://localhost:8000/admin/config/islandora/core
* Add a new namespace and corresponding rdf mapping 
* Clear cache
* Verify that jsonld has the expanded property uri (example: dbpedia-owl => http://dbpedia.org/ontology/, dbpedia-owl: notes, should result in:

```
http://dbpedia.org/ontology/description |  
@value | "sample notes"
@type | "http://www.w3.org/2001/XMLSchema#string"
```

# Interested parties
@whikloj @dannylamb

### DIFF
--- a/config/install/islandora.settings.yml
+++ b/config/install/islandora.settings.yml
@@ -3,3 +3,17 @@ fedora_rest_endpoint: 'http://localhost:8080/fcrepo/rest'
 _core:
   default_config_hash: nDZDR2rrpXXQ-D_7BYrdDFAXYOsB5hgH6vCAMV5I3w8
 broadcast_queue: islandora-connector-broadcast
+
+rdf_namespaces: >
+  ldp  => http://www.w3.org/ns/ldp#,
+  dc11 => http://purl.org/dc/elements/1.1/,
+  nfo => http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/v1.1/,
+  ebucore => http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#,
+  fedora => http://fedora.info/definitions/v4/repository#,
+  owl => http://www.w3.org/2002/07/owl#,
+  ore => http://www.openarchives.org/ore/terms/,
+  rdf => http://www.w3.org/1999/02/22-rdf-syntax-ns#,
+  islandora => http://islandora.ca/CLAW/,
+  pcdm => http://pcdm.org/models#,
+  use => http://pcdm.org/use#,
+  iana => http://www.iana.org/assignments/relation/,
--- a/islandora.module
+++ b/islandora.module
@@ -66,23 +66,18 @@ function islandora_modules_installed($modules) {
  * Implements hook_rdf_namespaces().
  */
 function islandora_rdf_namespaces() {
-  // Yes, it's amazing, rdf is not registered by default!
-  return [
-    'ldp'  => 'http://www.w3.org/ns/ldp#',
-    'dc11' => 'http://purl.org/dc/elements/1.1/',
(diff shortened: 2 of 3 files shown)


whikloj:
I would suggest that a table would be nicer for displaying the prefixes than a big text box. I've not done this yet, but you could put it on its own tab for more space. Then you could have a form with prefix and URL boxes.

You'll also want to verify the URL and/or prefix is already registered.

Also less prone to typos crashing the system. 

I'm wondering about making the settings.yaml parse into the actual array and then not have to split on the `=>`?

Much of this is just theory, but if you did 
```
rdf_namespaces:
  - ldp : http://www.w3.org/ns/ldp#
  - dc11 : http://purl.org/dc/elements/1.1/
  - nfo : http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/v1.1/
  - ebucore : http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#
  - fedora : http://fedora.info/definitions/v4/repository#
 ....
```
would that parse directly into the array with appropriate key => value pairs? Then you could just pass that back and forth.

Natkeeran:
@whikloj 

Good points.  I'll take these into consideration and refactor it.  

May I know how we can verify URL and/or prefix already register?

Also, how to remove/update prefixes in http://localhost:8080/fcrepo/rest/apix/extensions?

whikloj:
@Natkeeran you can verify what is already registered by calling `$array = rdf_get_namespaces()` and comparing. You can do a simple `array_key_exists($prefix, $array)` for the prefixes and do an `$flipped_array = array_flip($array)` then an `array_key_exists($url, $flipped_array)` to check for the URLs.

I don't think we register any RDF prefixes with API-X...but @dannylamb would know for sure.

Natkeeran:
Closing this for now.  Will come back to this later.

PR #82, opened 2018-03-16
Fix OOM composer errors

whikloj:
Islandora is crashy in Travis

Uses php without a memory limit to allow more memory to composer.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,15 @@ branches:
 before_install:
   - export SCRIPT_DIR=$HOME/CLAW/.scripts
   - export DRUPAL_DIR=/opt/drupal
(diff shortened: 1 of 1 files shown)


whikloj:
@Islandora-CLAW/committers 

I did a test with adding `memory_limit = -1` to all the php configs in the setup from CLAW ([commit](https://github.com/whikloj/CLAW/commit/263ae362af639da3e7a5e8e93ced21d875d19f91))

But then 5.6 still seems to die (https://travis-ci.org/whikloj/claw-jsonld/builds/354507558), so I guess we just need to add `php -dmemory_limit=-1` to every composer we run.

This seems to fix up Islandora, I'll add a similar PR to the other repos.

PR #83, opened 2018-03-20
Derivatives

dannylamb:
# DO NOT MERGE: NEEDS TESTS

**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/766

Provides the ability to trigger derivatives from entity update hooks.

# What's new?

* Lots of hook implementations in the .module
* A utility file with functions for derivatives
* Updated the service for adding and updating media
* Touched up the EmitEvent base class to pass configuration as a JSON attachment to the event.
* Added a DerivativeReaction

# How should this be tested?

Testing instructions to come.

# Interested parties


islandora.module
```diff
@@ -84,104 +87,171 @@ function islandora_rdf_namespaces() {
...
-      $reaction->execute($entity);
-    }
+  // Fire off derivative reactions for the Media.
+  foreach ($utils->getReferencingNodes($media->id()) as $node) {
+    $utils->executeDerivativeReactions(
+      '\Drupal\islandora\Plugin\ContextReaction\DerivativeReaction',
+      $node,
+      $media
+    );
   }
```
whikloj: Why only fire off derivative reactions on `media_update` only and not `media_insert`? 

Also, does this not collide with the derivative media reactions that occur in `node_update` because presumably the media that hit this hook will also be referenced by a node and hit that hook..

islandora.module
```diff
@@ -84,104 +87,171 @@ function islandora_rdf_namespaces() {
...
+
+  // Execute derivative reactions
+  $media_entities = \Drupal::service('entity_type.manager')->getStorage('media')->loadMultiple($filtered);
+  foreach ($media_entities as $media) {
+    foreach ($utils->getReferencingNodes($media->id()) as $node) {
+      $utils->executeDerivativeReactions(
+        '\Drupal\islandora\Plugin\ContextReaction\DerivativeReaction',
+        $node,
+        $media
+      );
```
whikloj: Essentially the same question as above, why only on `update` (I'm assuming its a Drupalism) and is there not a collision when calling these derivative reactions on media from files and calling them directly in the `media_update`?

src/DerivativeUtils.php
```diff
@@ -0,0 +1,379 @@
...
+
+  public function haveFieldsChanged(ContentEntityInterface $entity, ContentEntityInterface $original) {
+
+    $field_definitions = $this->entityFieldManager->getFieldDefinitions($entity->getEntityTypeId(), $entity->bundle());
+
+    $ignore_list = ['vid', 'changed', 'path'];
+    foreach ($field_definitions as $field_name => $field_definition) {
+      if (in_array($field_name, $ignore_list)) {
+        continue;
+      }
```
whikloj: Just wondering if it would be more efficient to never process the `$ignore_list`... something like.
```
$ignore_list = ['vid' => 1, 'changed' => 1, 'path' => 1];
$field_definitions = array_diff_keys($field_definitions, $ignore_list);
foreach ($field_definitions as $field_name => $field_definition) {
```

src/EventGenerator/EmitEvent.php
```diff
@@ -126,21 +126,13 @@ public function execute($entity = NULL) {
...
-    }
-    elseif ($this->configuration['event'] == 'delete') {
-      $message = $this->eventGenerator->generateDeleteEvent($entity, $user);
-    }
-
-    // Transform message from string into a proper message object.
-    $message = new Message($message, ['Authorization' => "Bearer $token"]);
+    // Generate event as stomp message.
+    $user = $this->entityTypeManager->getStorage('user')->load($this->account->id());
+dsm("Sending message to {$this->configuration['queue']}: {$this->eventGenerator->generateEvent($entity, $user, $this->configuration)}");
```
whikloj: Left over debug message

islandora.module
dannylamb: It's the direction of the relationships.  The only `create` hook I implement is for Nodes because they're at the top of the Node -> Media -> File hierarchy.  So create or update one of those and I need to traverse down and check for derivatives every time, even if they possibly don't reference anything yet.  But even with Inline Entity Form, Media and Files are created without being referenced by a Node, so the hierarchy is not intact and we can't traverse back up to find where to put derivatives.  So it's just updates for the Media and Files when everything is referenced.  Checking on node create and update catches all the other cases.

Embarassingly, I've left a dsm in here that prints the events every time they're sent to a queue to confirm this.  It's also essentially what the functional tests for this will look like, to confirm that we fire when we're supposed to and only once.

src/EventGenerator/EmitEvent.php
dannylamb: Haha, you found it. :+1:

src/DerivativeUtils.php
dannylamb: probably :man_shrugging:

dannylamb: definitely

dannylamb:
Wow... so... I just had a blast fighting with Drupal :fist_oncoming: 

Turns out when you're building up bundles and fields programatically in a test's `setUp()` method, not all the hooks fire / events respond.  So it was making the bundles and fields, but not generating the form and view modes.  So I couldn't set the form to display a media reference field in order to test it.  Any attempts to create form and view modes programatically failed.

The only thing that worked was loading them up from yaml files.  _BUT_ you can't go 'all the way' and just provide all your bundles and fields as yaml files.  It won't make the database tables for your fields!

So, TL;DR build your bundles and fields programatically, but load your other config afterwards using the [ConfigImporter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21ConfigImporter.php/class/ConfigImporter/8.2.x) that you can access from any test with `$this->configImporter()`

Anyway... thanks Drupal.  It's been real :sob:

dannylamb:
@Islandora-CLAW/committers Ok, this one's ready.  Two more to go and then I'll write up the test instructions.

dannylamb:
Superseded by https://github.com/Islandora-CLAW/islandora/pull/86

PR #85, opened 2018-04-06
View and form alters

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/821

* Depends on https://github.com/Islandora-CLAW/islandora/pull/83 (Needs it to load MOAR configuration to properly test that view and form modes are swapped)

# What does this Pull Request do?

Provides two new context reaction plugins: `ViewModeAlterReaction` and `FormDisplayAlterReaction`, and some logic in `islandora.module` to use them in response to `hook_entity_view_mode_alter()` and `hook_entity_form_display_alter()`.

# What's new?
It's all wrapped up with the Derivatives code because I needed the improvements to our test gear from it, so here's the salient diff:
```diff
diff --git a/islandora.module b/islandora.module
index 37112b8..c15551d 100644
--- a/islandora.module
+++ b/islandora.module
@@ -232,3 +232,59 @@ function islandora_jsonld_alter_normalized_array(EntityInterface $entity, array
     $reaction->execute($entity, $normalized, $context);
   }
 }
+
+/**
+ * Implements hook_entity_view_mode_alter().
+ */
+function islandora_entity_view_mode_alter(&$view_mode, EntityInterface $entity) {
+  $storage = \Drupal::service('entity_type.manager')->getStorage('entity_view_mode');
+  $context_manager = \Drupal::service('context.manager');
+
+  // Alter view mode based on context.
+  foreach ($context_manager->getActiveReactions('\Drupal\islandora\Plugin\ContextReaction\ViewModeAlterReaction') as $reaction) {
+    // Construct the new view mode's machine name.
+    $entity_type = $entity->getEntityTypeId();
+    $mode = $reaction->execute();
+    $machine_name = "$entity_type.$mode";
+
+    // Try to load it.
+    $new_mode = $storage->load($machine_name);
+
+    // If successful, alter the view mode.
+    if ($new_mode) {
+      $view_mode = $mode;
+    }
+    else {
+      \Drupal::logger('islandora')->info("EntityViewMode $machine_name does not exist.  View mode cannot be altered.");
+    }
+  }
+}
+
+/**
+ * Implements hook_entity_form_display_alter().
+ */
+function islandora_entity_form_display_alter(&$form_display, $context) {
+  $storage = \Drupal::service('entity_type.manager')->getStorage('entity_form_display');
+  $context_manager = \Drupal::service('context.manager');
+
+  // Alter form display based on context.
[...]

PR #86, opened 2018-05-02
Content modeling overhaul

dannylamb:
**GitHub Issues**:  Supersedes https://github.com/Islandora-CLAW/islandora/pull/83 and https://github.com/Islandora-CLAW/islandora/pull/85

- Resolves https://github.com/Islandora-CLAW/CLAW/issues/817
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/812
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/799
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/788
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/785
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/821
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/766
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/721
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/716
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/685
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/678
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/676
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/669
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/656
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/452

yyyyyeah... I think I've surpassed even @ruebot in conflating PRs/issues

Total content modeling overhaul, heavily based on taxonomy terms and the context module. 

# What's new?

* Just one content type
  * We use one content type (Repository Item), though it is neither a requirement nor restriction.  Just simplifies managing metadata.  The single type basically serves as a default metadata profile.
* Taxonomy Terms
  * Content and Media are tagged with taxonomy terms to denote what they represent (images, audio, video, etc...), what type of file they are (presevation master vs thumbnail, etc..) and general behavior (tagging content with 'OpenSeadragon' makes it display its contents using the OpenSeadragon viewer).
  * Terms are indexed in Fedora and are given an external URI that's used as its identifier instead of the local integer that gets minted.  All configuration that's exported that references terms exports them as these URIs, and Islandora does the lookups where appropriate.  
* Features
  * Using features to package configuration provides a better user/developer experience when working with lots of config files.
* Migrate
  * Uses migrate_plus, migrate_tools, and migrate_source_csv to provide default taxonomy terms for the module.  Because the migrations are named and are part of the 'islandora' migration group, the terms can be referenced in your own migrations by their external uri.
* Views
  * Using views to manage members and media, with action links and bulk update operations.
* Core Media
  * Using the Core media module instead of third-party media_entity module.
  * We now have audio, video, and binary for free!  Pdfs even work ok as binaries because clicking on the download link usually opens pdf.js in your browser.
  * References are now reversed (e.g. Media reference their parent nodes, not nodes referencing their child media).
  * Taxonomy terms are used to indicate what 'type' it is (preservation master, service file, thumbnail, etc..)
* Switch view mode as a context reaction
* Switch form mode as a context reaction
* Image derivatives
  * Another Drupal Action wrapped in a ContextReaction (like indexing and unindexing)
  * Totally customizable through the core Config -> Actions screen.
  * Asynchronous and distributed, works with Alpaca and Crayfish to generate derivatives in a scalable fashion.
* Link Headers
  * Referenced taxonomy terms are added to responses as link headers using the rel=tag relationship.
* Idempotent media updates
  * You can update Media via REST idempotently (was POST, now its PUT).  The endpoint for doing so is now `/node/{node}/media/{media_type}/{taxonomy_term}`.  
# How should this be tested?

## Use claw-playbook
Pull in the changes from https://github.com/Islandora-Devops/claw-playbook/pull/66 into your claw-playbook repo, or clone down my branch directly `git clone -b content-modeling-overhaul https://github.com/dannylamb/claw-playbook.git`

Then provision a new box (sry... gotta destroy your existing env first), log in, and pipe down a bash script to finish the setup.
- `vagrant up`
- `vagrant ssh`
- `curl https://gist.githubusercontent.com/dannylamb/99f8720faaa20c37e30454c88468dd1a/raw/e8fdfdc6225a2ef208527e09cab2db05af6e5b16/gistfile1.txt | sudo bash`

Now you can play around with all the features.  Screenshots and examples of what you can do to come.

# Interested parties
@Islandora-CLAW/committers


dannylamb:
Tests are going to fail here.  I'm setting up the other PRs and will circle back.

modules/islandora_feature/config/install/rdf.mapping.node.islandora_object.yml
```diff
@@ -0,0 +1,41 @@
...
+id: node.islandora_object
+targetEntityType: node
+bundle: islandora_object
+types:
+  - 'pcdm:Object'
+fieldMappings:
+  field_description:
+    properties:
+      - 'dc:description'
+  field_memberof:
```
seth-shaw-unlv: This field is now "field_member_of".

seth-shaw-unlv:
Add Image isn't allowing Tiffs.

screen shot 2018-05-03 at 8 51 02 am

seth-shaw-unlv:
@dannylamb Yeah, I realized what was happening after I submitted the comment. So far everything is looking good in my testing.

I'm almost done with creating a new metadata profile content type with Creator and Subject entity references (from my existing migration example) that has the same functionality as Repository Item. I just need to get the updated yml files in my test module so I can start updating the migration itself.

seth-shaw-unlv:
Okay, my [CLAW migration proof-of-concept](https://github.com/seth-shaw-unlv/claw-migrate-files-poc) now uses the updated content type overhaul. Note that the migration does not trigger actions, so those need to be done manually after the items are loaded. (See step 5 the instructions in my README.)

ANYWAY, this PR looks good to me.

seth-shaw-unlv:
Still trying to get migrated items into Fedora/triplestore and I came across this issue: when adding items using the web interface Milliner isn't happy:

`app.ERROR:  {"Exception":"[object] (RuntimeException(code: 500): Cannot parse file UUID from http://localhost:8000/media/26?_format=json.  Media must use 'field_file' or 'field_image'. at /var/www/html/Crayfish/Milliner/src/Service/MillinerService.php:368)"} []`

It appears that Milliner is expecting field_file or field_image while the jsonld is providing field_media_file and field_media_image instead.

dannylamb:
@seth-shaw-unlv Nice.  Good catch.  Looks like I've got something hard-coded in milliner.

dannylamb:
So this has tests passing and coding standards ok for me locally.

... _waits for Travis_ ...

dannylamb:
...caught by copy paste detector in my tests :cry:

dannylamb:
Yassss... Travis is appeased.

dannylamb:
Ok, I just tested a fresh `vagrant up` and ran the updated gist.  Events are firing during migrations!!!

Thanks @seth-shaw-unlv

dannylamb:
I'll update that shortly.  Thanks @seth-shaw-unlv

dannylamb:
Here's some things worth trying from a fresh install:

## Make sure our terms made it to Fedora
- This new approach to content modeling relies heavily on Drupal's taxonomy module.  We tag content with taxonomy terms, and use that to control the site's behavior.  We can do things like switch view and form modes, restrict access, and denote what the repository item represents (image, binary, audio, video) using these tags.  They're super important, and in conjunction with the `context` module, you can do just about anything with them.
- So to verify they're in Fedora, go to http://localhost:8080/fcrepo/rest and verity that you've got some taxonomy terms in your Fedora
screenshot from 2018-05-10 11-34-17
- Click on one and you should see it has an owl:sameAs predicate pointing to an external uri
screenshot from 2018-05-10 12-20-50

## Make a collection
- Go to http://localhost:8000 and log in with user `admin` and password `islandora`.
- Click Add Content > Repository Item and fill out the form. Give it a tag of 'Collection`.
screenshot from 2018-05-10 11-35-32

## Add an Image to the collection
- From your collection, click the Members tab > Add member -> Repository Item and fill out the form.  Note that the member of field is pre-filled.  Give it a tag of 'Image'.
screenshot from 2018-05-10 12-40-54
- From your image's page, click the Media tab > Add media > Image and fill out the form.  Note that the media of field is pre-filled.  Upload an image and give it a tag of "Preservation Master".
screenshot from 2018-05-10 12-45-39
- You'll be redirected to the media's page.  Notice that some very basic technical metadata is extracted, like mimetype, file size, image width/height.
screenshot from 2018-05-10 16-54-06
- Click on the 'Media Of' link to get sent back to the node's page.  You should see the image on the Node's page, too.  It's actually a view that's rendering the "Service File" derivative that got generated (more on that later).  Click on its 'Media' tab and you should now see all the derivatives that were generated.
screenshot from 2018-05-10 16-56-14
## Delete a media and its file
- There's two new actions available to you for deleting media.  You can delete media and leave the files alone, or you can delete a media _and_ its associated file(s).  You can access these from any view that has bulk operations enabled (and anywhere in code as well).  For example, if we're still on the view to manage media for our basic image, you can select 'Delete media and file(s)' from the dropdown and torch the service file that was generated.
screenshot from 2018-05-10 16-57-46
- Go back to the node and you shouldn't see the image getting rendered anymore.  Without the service file, the view that's getting rendered out has no results.
screenshot from 2018-05-10 17-04-10
- If you want to regenerate that derivative, go to Admin > Content, select "Generate a service file from a preservation master".  Then click the checkbox for your image's node and hit "Apply to selected items" .
screenshot from 2018-05-10 17-04-37
- Boom... your derivative is regenerated.  Go visit the page again, and the image should be back.
screenshot from 2018-05-10 17-07-14
- Go to its 'Media' tab and you should be able to see the service file there, too.

## Make a 'Large Image'
- Go to `http://localhost:8000/admin/content` and click 'Add content' > Repository Item.  Fill out the form to create the node.  You can make it a member of your collection using the autocomplete field for 'Member of'.  Give it the tags 'Image' and 'Openseadragon'.
screenshot from 2018-05-11 14-39-52
- Drupal thinks tiffs are files, not images.  So click on your new node's 'Media' tab and then 'Add media' > File.  Upload a tiff and give it the preservation master tag just like we did with the basic image.
screenshot from 2018-05-11 14-46-04

-Save it, and derivatives will be generated.  Follow the 'Media of' link to go back to your node and there you have it, a 'Large' image using the Openseadragon viewer. 
screenshot from 2018-05-11 14-47-09


- So both the 'basic' and 'large' images are nodes that have associated media.  And ingesting media tagged with 'Preservation Master' triggers derivatives.  The only real difference between the two is that the 'Large' image is tagged wtih 'Openseadragon', signaling that we want to use the Openseadragon viewer.   So what happens if we remove the Openseadragon tag?
screenshot from 2018-05-11 14-50-45
- It reverts to showing the Service File just like 'basic' image!  But wait, there's more!  Remove the 'Image' tag on the node and replace it with 'Binary'.
screenshot from 2018-05-11 14-51-45
- Now it treats the tiff as a regular, non-image file and offers you a download link.

## Audio and Video
- We aren't generating audio or video derivatives yet, but you can still ingest them into Islandora and display them.  Go to `admin/content` and click 'Add content' > Repository Item.  Fill out the form and tag th object as 'Video'.
screenshot from 2018-05-11 15-01-47

- Click on your node's 'Media' tab and click 'Add Media' > 'Video'.  Upload your video and give it a tag of 'Service File'.  The default display renders service files to the screen, so by tagging it as such we'll see its viewer when we go visit the node.
screenshot from 2018-05-11 15-02-54
screenshot from 2018-05-11 15-03-20
 - You just get the html5 viewer, but hey, it works.  You can repeat the process with an audio file and you'll get the html5 audio player.  Just remember to tag the media as a service file.

## Context
So how does this all work?  It's the `context` module, which lets you use Drupal condition plugins to define when actions will be triggered.  We're using it to do everything from index in Fedora and the triplestore to generating derivatives to switching view modes and even themes! Check it out at `admin/structure/context`.
screenshot from 2018-05-11 15-07-58
- Want to only index certain files in Fedora?  Go for it.  Want to give certain users a simpler form?  No problem.  Want to have complete control over when and how derivatives fire?  By combining Conditions and Actions, you can do just about anything through a UI.  For example, click on the 'Content' context and you'll see we're indexing all 'Repository Item' nodes in Fedora and the triplestore.
screenshot from 2018-05-11 15-14-37
- You can easily flush other content types to Fedora.  Or even cooler, you could add a 'Node has term' condition, to make only 'Images' get indexed.
screenshot from 2018-05-11 15-16-02
- It's a bit of a silly example, but it just goes to show you how powerful this all is behind the scenes.  Lots of functionality that was previously hidden in code in hooks and alters can now be exposed to users via forms!

ajs6f:
Arrrg! I have to get a working install done before I can enjoy all this tasty goodness.

Natkeeran:
@dannylamb 

I was able to follow your instructions closely.  Wow.  Looks good.

Few minor issues/notes:
* tags are in drupal, but not in fedora
* deleting and adding a media tagged preservation does not seem to trigger the derivative generation automatically
* Member of shows all nodes, not just nodes that are tagged as collections

seth-shaw-unlv:
@dannylamb the installation instructions still use [the old gist](https://gist.githubusercontent.com/dannylamb/99f8720faaa20c37e30454c88468dd1a/raw/0c16e35979fe03a38d2057f7a4980bdc316cc813/gistfile1.txt), not [the new one](https://gist.githubusercontent.com/dannylamb/99f8720faaa20c37e30454c88468dd1a/raw/1e5d33cc97360ae2895e796f49d097285de2688e/gistfile1.txt). That might account for @Natkeeran's missing tags.

dannylamb:
@Natkeeran @seth-shaw-unlv Ah, ok.  I thought it was the same url but i guess different revisions get their own url.  Let me update those.

dannylamb:
Yeah... that's a different URL alright.  Sorry for the confusion @Natkeeran

ppound:
Followed the instructions above, things went smoothly and as far as I can tell everything I tried so far is working as expected.

mjordan:
What @ppound said. Nice work! One question: the filename assigned to the image derivatives uses the UUID of the , e.g. `2683c9cb-a9d2-47b8-8c11-b2c515517bb6 - ServiceFile.jpeg

mjordan:
Added my own term to the taxonomy and it showed up in fcrepo as expected. But I noticed that there is HTML markup in the description:

taxo

Just thought I'd point that out. Don't want to end up with a similar issue to https://jira.duraspace.org/browse/ISLANDORA-2185. Should I open a separate issue for this?

mjordan:
And for the record, taxonomy terms are referenced in the response body to REST requests (see bottom of response):

```
curl -v -u "admin:islandora" "http://localhost:8000/node/4?_format=json"
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
* Server auth using Basic with user 'admin'
> GET /node/4?_format=json HTTP/1.1
> Authorization: Basic YWRtaW46aXNsYW5kb3Jh
> User-Agent: curl/7.35.0
> Host: localhost:8000
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Tue, 15 May 2018 13:36:50 GMT
* Server Apache/2.4.18 (Ubuntu) is not blacklisted
< Server: Apache/2.4.18 (Ubuntu)
< X-Powered-By: PHP/7.0.30-0ubuntu0.16.04.1
< Cache-Control: must-revalidate, no-cache, private
< Link: <http://localhost:8000/node/4>; rel="canonical"
< Link: <http://localhost:8000/node/4/delete>; rel="https://drupal.org/link-relations/delete-form"
< Link: <http://localhost:8000/node/4/edit>; rel="edit-form"
< Link: <http://localhost:8000/node/4/revisions>; rel="version-history"
< Link: <http://localhost:8000/node/4>; rel="https://drupal.org/link-relations/revision"
< Link: <http://localhost:8000/node?node=4>; rel="https://drupal.org/link-relations/create"
< Link: <http://localhost:8000/node/1>; rel="related"; title="Member of"
< Link: <http://purl.org/coar/resource_type/c_c513>; rel="tag"; title="Image"
< Link: <http://localhost:8000/media/6>; rel="related"; title="Preservation Master"
< Link: <http://localhost:8000/media/7>; rel="related"; title="Service File"
< Link: <http://localhost:8000/media/8>; rel="related"; title="Thumbnail"
< Link: <http://localhost:8000/node/4?_format=jsonld>; rel="alternate"; type="application/ld+json"
< X-Drupal-Dynamic-Cache: HIT
< X-UA-Compatible: IE=edge
< Content-language: en
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Expires: Sun, 19 Nov 1978 05:00:00 GMT
< Vary: 
< X-Generator: Drupal 8 (https://www.drupal.org)
< Content-Length: 1290
< Content-Type: application/json
< 
{"nid":[{"value":4}],"uuid":[{"value":"2683c9cb-a9d2-47b8-8c11-b2c515517bb6"}],"vid":[{"value":5}],"langcode":[{"value":"en"}],"type":[{"target_id":"islandora_object","target_type":"node_type","target_uuid":"f01492af-69ff-4d30-a71b-4469ac4d2d74"}],"revision_timestamp":[{"value":"2018-05-14T22:47:09+00:00","format":"Y-m-d\\TH:i:sP"}],"revision_uid":[{"target_id":1,"target_type":"user","target_uuid":"00f14965-3396-42f5-b6bc-71c922de2d6a","url":"\/user\/1"}],"revision_log":[],"status":[{"value":true}],"title":[{"value":"Second imate"}],"uid":[{"target_id":1,"target_type":"user","target_uuid":"00f14965-3396-42f5-b6bc-71c922de2d6a","url":"\/user\/1"}],"created":[{"value":"2018-05-14T22:46:57+00:00","format":"Y-m-d\\TH:i:sP"}],"changed":[{"value":"2018-05-14T22:47:09+00:00","format":"Y-m-d\\TH:i:sP"}],"promote":[{"value":true}],"sticky":[{"value":false}],"default_langcode":[{"value":true}],"revision_translation_affected":[{"value":true}],"path":[{"alias":null,"pid":null,"langcode":"en"}],"field_description":[],"fie
ld_member_of":[{"target_id":1,"target_type":"node","target_uuid":"15b855a1-61e4-4071-8142-9d9140e31085","url":"\/node\/1"}],"field_tags":[{"target_id":8,"target_type":"taxonomy_term","target_uuid":"a84cc404-18be-405a-bd5a-3fe55f3af520","url":"\/taxonomy\/term\/8"}]}
```

and you can then request info on the term:

```
curl -v -u "admin:islandora" "http://localhost:8000/taxonomy/term/8?_format=json"
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
* Server auth using Basic with user 'admin'
> GET /taxonomy/term/8?_format=json HTTP/1.1
> Authorization: Basic YWRtaW46aXNsYW5kb3Jh
> User-Agent: curl/7.35.0
> Host: localhost:8000
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Tue, 15 May 2018 14:03:52 GMT
* Server Apache/2.4.18 (Ubuntu) is not blacklisted
< Server: Apache/2.4.18 (Ubuntu)
< X-Powered-By: PHP/7.0.30-0ubuntu0.16.04.1
< Cache-Control: must-revalidate, no-cache, private
< Link: <http://localhost:8000/taxonomy/term/8>; rel="canonical"
< Link: <http://localhost:8000/taxonomy/term/8/delete>; rel="https://drupal.org/link-relations/delete-form"
< Link: <http://localhost:8000/taxonomy/term/8/edit>; rel="edit-form"
< Link: <http://localhost:8000/taxonomy/term?taxonomy_term=8>; rel="https://drupal.org/link-relations/create"
< X-Drupal-Dynamic-Cache: MISS
< X-UA-Compatible: IE=edge
< Content-language: en
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Expires: Sun, 19 Nov 1978 05:00:00 GMT
< Vary: 
< X-Generator: Drupal 8 (https://www.drupal.org)
< Content-Length: 828
< Content-Type: application/json
< 
* Connection #0 to host localhost left intact
{"tid":[{"value":8}],"uuid":[{"value":"a84cc404-18be-405a-bd5a-3fe55f3af520"}],"langcode":[{"value":"en"}],"vid":[{"target_id":"tags","target_type":"taxonomy_vocabulary","target_uuid":"ef592ce6-bbe0-4fc8-8eb3-e93b452d980b"}],"name":[{"value":"Image"}],"description":[{"value":"A visual representation other than text, including all types of moving image and still image","format":null,"processed":"\u003Cp\u003EA visual representation other than text, including all types of moving image and still image\u003C\/p\u003E\n"}],"weight":[{"value":0}],"parent":[],"changed":[{"value":"2018-05-14T18:47:28+00:00","format":"Y-m-d\\TH:i:sP"}],"default_langcode":[{"value":true}],"path":[{"alias":null,"pid":null,"langcode":"en"}],"field_external_uri":[{"uri":"http:\/\/purl.org\/coar\/resource_type\/c_c513","title":null,"options":[]}]}
```

Neat!

seth-shaw-unlv:
This PR is breaking block placement. I attempted to place a block on the main content area but save button would not work until I had selected tags in each of the "has term" tabs (which didn't apply in this case).

screen shot 2018-05-15 at 11 39 12 am

Even if I entered a tag in each *required* field and selected "Negate" the block still did not appear.

DiegoPino:
@seth-shaw-unlv probably the ::evaluate() method for the new Condition Plugins and the checks that are being done are "blocking the blocks."

This is a busy pull and i can't follow the whole 8000 new lines nor i fully understand the whole architecture here (i get the Condition Plugins but i don't understand how the chain evaluation is or why terms are forced in blocks form) but lookig at 
https://www.drupal.org/node/2287827
 and 

https://github.com/dannylamb/islandora/blob/c2935bd79b3876a435a03e0a2543423d69f2a9ef/src/Plugin/Condition/NodeHasTerm.php#L20
or this
https://github.com/dannylamb/islandora/blob/c2935bd79b3876a435a03e0a2543423d69f2a9ef/src/Plugin/Condition/ParentNodeHasTerm.php

Seems to me that ::evaluate logic is interfering with blocks and could with other cases too.

@Islandora-CLAW/committers is there a non functional( means testing the UI) way of reviewing this pull someone could suggest? Like a wire diagram of all the new moving pieces?

DiegoPino:
@seth-shaw-unlv maybe here? https://github.com/dannylamb/islandora/blob/c2935bd79b3876a435a03e0a2543423d69f2a9ef/src/Plugin/Condition/NodeHasTerm.php#L136 😰 

I know the issue is in the Conditional Plugin 😸  evaluation piece (the exact where NOT!)  and to me, it seems like that place where entities (any?) are evaluated `::evaluateEntity()` and checked for every defined term presence, could be the culprit?

But to the honest. I would need to deploy the whole machinery and add debug statements. I don't know enough!

src/Plugin/Condition/NodeHasTerm.php
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['term'] = [
+      '#type' => 'entity_autocomplete',
+      '#title' => $this->t('Term'),
+      '#tags' => TRUE,
+      '#required' => TRUE,
+      '#default_value' => $this->utils->getTermForUri($this->configuration['uri']),
```
seth-shaw-unlv: Drupal issues a notice when the configuration is missing the 'uri' index: `Notice: Undefined index: uri in Drupal\islandora\Plugin\Condition\NodeHasTerm->buildConfigurationForm() ...`

seth-shaw-unlv:
@dannylamb Dropping the requirement is the first step, but the [eval statements](https://github.com/Islandora-CLAW/islandora/pull/86#issuecomment-389294119) are [still problematic](https://github.com/Islandora-CLAW/islandora/pull/86#issuecomment-389300795). I put this topic on the agenda along with an (untested) idea for resolving it.

dannylamb:
@seth-shaw-unlv Aha.  Ok.  So there's a pattern that repeats throughout those core conditions that has always puzzled me.  Turns out, it's there to handle this situation.  In the `NodeType` condition there's
```php
    if (empty($this->configuration['bundles']) && !$this->isNegated()) {
      return TRUE;
    }
```
and in the `UserRole` condition it's
```php
    if (empty($this->configuration['roles']) && !$this->isNegated()) {
      return TRUE;
    }
```

Looks like I just need to add similar logic to the new conditions.  The simple ones like `IsNode` and `IsMedia` that don't have a form are going to need to have a checkbox in them so you can enable/disable and run a similar check, but hey.  Could be worse.

seth-shaw-unlv:
@dannylamb Or we could turn the binary ones into a single condition with checkboxes. That way we can check for the form.

dannylamb:
Ok, so in trying to write up a lot of documentation for this, I'm finding something that's probably best to fix now.  There's some config that _needs_ to be there for Islandora to work, and then there's some that is just _an example_ of how you can do things.  I needs to separate those out and ship with two features.  One that's sorta sacred, can't be touched, and must be enabled. And one that's more of a suggestion/example of what you can do.

DiegoPino:
@dannylamb, maybe a sprint effort to make the optionals into an islandora_demo module? Or too much separation?

dannylamb:
@DiegoPino Oh, that level of separation is what I'm going for.  I'll partition the config we have right now as best I can, and we'll go from there.  Basically, I need to be able to say in documentation "At a minimum, enable this core feature and you can do whatever content modeling you want.  If you want to see a working example, check out this other feature."  That's something I can do easily now, but will have to continuously re-evaluate as people spin up sites and try modeling content.

dannylamb:
Ok, i've done a pretty detailed README update.  Still needs more documenting, but there's a lot there that hadn't been put down in words yet.

I also combined all the config into two features: `islandora_core_feature` which is required for using Islandora.  And `islandora_demo_feature` which installs all the config so that you can do the testing outlined above.  That's as good of a split as I can make until we feedback from people trying it out and attempting to do their own thing.

dannylamb:
@seth-shaw-unlv Once travis is happy, let's give this one last sanity check and then you can do the honors.  I'll then update the crayfish pull and we can slice the versions for Crayfish and Alpaca.

ajs6f:
Ok, finally ready to try to help test this. Are the original instructions for getting it up and running at the very top of this page still good? @dannylamb does your `-playbook` branch have the various fixes for CentOS incorporated?

ajs6f:
Wait, just realized that testing is pretty much over. :sigh:

dannylamb:
@ajs6f You testing this is still worthwhile.  Especially if it's on CENTOS.

In theory, it should all pick up the changes since it was in an external role.  I don't think I need to update my claw-playbook branch for that.  @seth-shaw-unlv are you on CENTOS when testing?

ajs6f:
Ok, @dannylamb , I'll give it a shot! Since I'm rolling onto bare metal, what exactly does "Then provision a new box (sry... gotta destroy your existing env first)" mean? Presumably that's just for Vagrant usage?

ajs6f:
Hm, fresh install isn't really very easy-- I would have to get another group to blow away the box and reimage. That would be several weeks of waiting. Given that we have zero content (well, I uploaded a test image for fun), can anyone see any pits into which I will fall by trying to load @dannylamb 's work over a "stock" `-playbook` install?

seth-shaw-unlv:
@ajs6f a stock claw-playbook with CentOS doesn't build. I have a setup that works but we have been slow getting the PRs in.

ajs6f:
`eats words`
Sorry, that wasn't well-phrased, I just meant "stock" as opposed to @dannylamb's work. IOW, I now have an install on CentOS bare metal, and I want to test this work by redeploying directly over that, and I'm trying to understand if the advice above (blow away before trying) refers only to content (of which I have none) or config/code as well.

ajs6f:
Foo. Then I have to take back my intention of testing-- it will take weeks to reset the box. I'll wait until this can be an upgrade that doesn't require blowing away the box. Thanks for clarifying, @seth-shaw-unlv !

whikloj:
Coder violation for PHP 7.x and OOM for 5.6. Maybe we should add some additional setup so composer is always using the PHP version and not the Ruby stub.

dannylamb:
Nooooooooooooo.........

Thanks @whikloj

dannylamb:
Well... there's the fix for coder.  But the OOM error is concerning, we already have the memory limit code in .travis.yml....

dannylamb:
I'm taking shots in the dark, here.  @whikloj, can you break this down for me?
> Maybe we should add some additional setup so composer is always using the PHP version and not the Ruby stub.
I'm willing to try anything.

Other than that... this is a compelling reason to ditch 5.6, which is just going to continue to cause problems as time goes on.

whikloj:
@dannylamb So if you type `composer` in a Travis box you get a Ruby stub command,  it does not seem to behave like real composer.

I had copied down a Travis docker container when I was trying this out and running `composer` by itself seemed to throw a weird Ruby message at me.

Anyways my thought was something more like https://github.com/Islandora-CLAW/CLAW/pull/818/files#diff-462959ea16db01e1b01c779895eaaacb check if we know where composer is and then call it directly with php.

dannylamb:
@whikloj you know, i was just in there thinking that might be the way to go.  i was wondering if putting it in the ini for 5.6 would be the way to go.

whikloj:
@dannylamb I don't know that putting it in the `.ini` has any difference. But if we do then we really need to call composer with php or (again this is my impression) it is not respecting that setting.

dannylamb:
@whikloj Well, let's see what happens.  But yeah, i think it's not respecting it at all.  According to Travis docs, there's 7.5 GB available, but we're popping at 1.5GB.

whikloj:
@dannylamb doing some reading, we don't store the `composer.lock` in `drupal-project` maybe we should, then in the `travis_setup_drupal.sh` we run
```
composer drupal-scaffold
composer install
```

whikloj:
We can even make multiple copies of the lock file (for each PHP version) and copy the correct one out.

dannylamb:
@whikloj sounds like it's worth trying.  i know when committing the lock file we ran into all kinds of issues, which is why we ended up removing it.  then again, we never ran `composer drupal-scaffold` either.  i'll give it a shot.

dannylamb:
@whikloj ^^ PR for those install changes

dannylamb:
qq

dannylamb:
@whikloj Autoloading the scaffolding script bombs.  I'm noticing some discrepancies with our fork of drupal-project and upstream that have to do with autoloading.  Let me see if I can pull something together so we can test that.

whikloj:
Yeah I saw that, I added a PR to CLAW. But hold on and I'll test running `composer require drupal-composer/drupal-scaffold:dev-master` first

dannylamb:
So I managed to set up `drupal-scaffold` as a post-install and post-update script, so there's no need to call them manually anymore.  I managed to do it locally without issue, but it blew up because I need to re-generate the lockfile using a 5.6 environment. The new upstream changes add a new dependency no in our lockfile currently.

whikloj:
Hmmm we already include drupal-scaffold. I think it is time to consider not fully re-installing Drupal for each build. The Drupal Scaffold docs indicate committing the scaffold files once you have change the version of Drupal.

So instead of building from drupal-project each time, we build a base Drupal and pull it down from a new repo?

whikloj:
@dannylamb I have a VM with PHP 5.6 on it I can use to generate a lock file if I can access your work.

ajs6f:
> I think it is time to consider not fully re-installing Drupal for each build

➕ 1000

dannylamb:
@whikloj Ok.  hang on.

@ajs6f I think the command is smart enough to know it's already been installed.  I don't think it's actually a full re-install every time.  But it would be prudent to check...

ajs6f:
@dhlamb Cool. My target (for SI) is a deployment that consists entirely of copying files into their final locations, setting perms, and a restart of the HTTPd service. I realize that the vast majority of CLAW's sites won't be under constraints as strict as SI's, so consider that a fun benchmark-- if it works for us, it should work for _anyone_.

dannylamb:
@whikloj https://github.com/dannylamb/drupal-project/tree/merge-from-upstream

whikloj:
@dannylamb it appears like drupal-scaffold brings in files needed before the install can happen. So maybe they should be in a pre-install section? Or we can run it explicitly.

dannylamb:
@whikloj i was under the impression that the install has to happen before, otherwise drupal-scaffold can't autoload.

dannylamb:
either way, i think running explicitly is the way to go

whikloj:
@dannylamb Ok, I'm not really sure but then we would still need the drupal-scaffold to execute if we run `composer install` instead of `composer update`. 

Because `drupal-scaffold` is called automatically (if you run `composer update`) to pull down the necessary files to build a Drupal when you change versions and because we don't commit any files to a stable store and instead have them downloaded everytime fresh. This is what I mean by "we install Drupal everytime", our process requires a composer update.

dannylamb:
@whikloj ok, let me issue a PR.  your lockfile works on 7.1 and is gtg if you ask me.  if you can update your PR on CLAW then we'll give this a whirl

whikloj:
@dannylamb Updated the CLAW PR.

dannylamb:
@whikloj merged Islandora-CLAW/CLAW#838

dannylamb:
Restarting Travis...

dannylamb:
I hope this works

whikloj:
Oh noes 😭

dannylamb:
Nooooooooooooooooooooooooooooooooo

dannylamb:
Yeah, this is solvable.

whikloj:
DOH!!!
https://github.com/Islandora-CLAW/CLAW/blob/master/.scripts/travis_setup_drupal.sh#L45

dannylamb:
gah, i've only been watching 5.6

dannylamb:
I'll restart the build, just looks like bad connections.

dannylamb:
......and timeout.

Restarting...

dannylamb:
Man... yeah... they either need to run in parallel or we have to collapse them all where possible, which sort of blurs the lines between all the tests.

ajs6f:
Here's to you

dannylamb:
@seth-shaw-unlv Think I've finally got the testing instructions re-re-updated.  I found a few things when enabling/disabling the features over and over again.  Nothing major, just stopped some config from uninstalling itself when getting rid of the demo feature.

I'm giving it another pass myself before shooting a demo video.  So I'll be giving it a lengthy run through again this afternoon.

seth-shaw-unlv:
Tested this morning with [the new gist](https://gist.githubusercontent.com/dannylamb/99f8720faaa20c37e30454c88468dd1a/raw/7d23441dfe3bfdeb0a0c2b6b1c2b30afb01c8058/gistfile1.txt) and it appears to be working fine.

dannylamb:
Keep trying to record a demo for @jonathangreen and I just keep finding stuff...  Hang on, more to come.

dannylamb:
Ok, added some basic validation to the media and membership views so you get 'Page Not Found' when no node exists with the supplied id.

whikloj:
I'm bringing this up again in claw-playbook (using Vagrant 2.1.1) for (hopefully) final testing.

whikloj:
That means no more commits @dannylamb

ajs6f:
Just one more thing...

dhlamb:
Just getting in some last minute changes from https://youtu.be/hCpktVpfCiw

whikloj:
@dannylamb Ok, let me know when you're good and I'll test it then

whikloj:
Oh you did, I just wanted to make sure there wasn't more work going on. 

I am now curl the bash gist, to use a generally illegible sentence.

whikloj:
Ok, derivatives are breaking Fedora. If you look at one of the generated files (Service or Thumbnail) you get a stack trace back from Fedora.
```
ubuntu@claw:/var/www/html/Crayfish/Gemini$ curl -i -H"Accept: text/n3" "http://localhost:8080/fcrepo/rest/70/80/71/78/70807178-cf92-404c-83f7-ee339e089526"
HTTP/1.1 500 Internal Server Error
Server: Apache-Coyote/1.1
ETag: "84c5f44b80d42f0a93d931b128cf11cc1f27d7ef"
Last-Modified: Mon, 04 Jun 2018 18:15:32 GMT
Content-Type: text/n3;charset=utf-8
Content-Length: 5150
Date: Mon, 04 Jun 2018 18:54:27 GMT
Connection: close

java.lang.IllegalArgumentException: Error parsing media type 'image/jpeg^^http://www.w3.org/1999/02/22-rdf-syntax-ns#langString^^en'
	at org.glassfish.jersey.message.internal.MediaTypeProvider.fromString(MediaTypeProvider.java:92)
	at org.glassfish.jersey.message.internal.MediaTypeProvider.fromString(MediaTypeProvider.java:60)
	at javax.ws.rs.core.MediaType.valueOf(MediaType.java:179)
	at org.fcrepo.http.api.FedoraLdp.getResource(FedoraLdp.java:240)
	at sun.reflect.GeneratedMethodAccessor35.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:161)
	at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:160)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:99)
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:389)
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:347)
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:102)
	at org.glassfish.jersey.server.ServerRuntime$2.run(ServerRuntime.java:326)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:271)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:267)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:267)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:317)
	at org.glassfish.jersey.server.ServerRuntime.process(ServerRuntime.java:305)
	at org.glassfish.jersey.server.ApplicationHandler.handle(ApplicationHandler.java:1154)
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:473)
	at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:427)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:388)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:341)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:228)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:292)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:207)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:240)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:207)
	at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:213)
	at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:106)
	at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:494)
	at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:141)
	at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:79)
	at org.apache.catalina.valves.AbstractAccessLogValve.invoke(AbstractAccessLogValve.java:616)
	at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:88)
	at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:522)
	at org.apache.coyote.http11.AbstractHttp11Processor.process(AbstractHttp11Processor.java:1095)
	at org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:672)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1520)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.run(NioEndpoint.java:1476)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.text.ParseException: Next event is not a Separator
	at org.glassfish.jersey.message.internal.HttpHeaderReader.nextSeparator(HttpHeaderReader.java:141)
	at org.glassfish.jersey.message.internal.HttpHeaderReader.readParameters(HttpHeaderReader.java:306)
	at org.glassfish.jersey.message.internal.HttpHeaderReader.readParameters(HttpHeaderReader.java:296)
	at org.glassfish.jersey.message.internal.MediaTypeProvider.valueOf(MediaTypeProvider.java:118)
	at org.glassfish.jersey.message.internal.MediaTypeProvider.fromString(MediaTypeProvider.java:90)
	... 49 more
```

This is a Fedora bug (as in Fedora shouldn't accept this in the first place), but before I can open a ticket there we should figure out how we are double categorizing the mimetype...
```
'image/jpeg^^http://www.w3.org/1999/02/22-rdf-syntax-ns#langString^^en'
```

dannylamb:
Whoa, ok crazy.  I suppose we shouldn't have the langstring at all if it's a mimetype.  But TBH I never anticipated this, lol.  Have you checked what you get if you try to pull it in via Drupal rest with `_format=jsonld`?

It has to be either the `jsonld` module or Milliner.

whikloj:
I am guessing it is `jsonld`, the media jsonld has this
```
      "http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#hasMimeType": [
        {
          "@value": "image/jpeg",
          "@language": "en"
        }
      ]
```

I'm guessing language is not needed as this isn't really a string.

whikloj:
But that doesn't explain the `^^http://www.w3.org/1999/02/22-rdf-syntax-ns#langString^^en`. At worst we should get a mime-type `image/jpeg^^en`. Which isn't exactly correct, but the above is crazy.

whikloj:
Unless this is entirely a Fedora problem and because of the internal use of the `ebucore:hasMimeType` it is munging it up.

dannylamb:
So, `ebucore:hasMimeType` is one of those magic properties that's totally relevant to us in CLAW, but just seems like a bad idea to enter a hard-coded check for it in the `jsonld` module.  So we either need to check if the field is translatable or do something specific in the alter we've already got in `islandora`

dannylamb:
Yeah I have no idea what's up with that URI getting shoved in there

whikloj:
Okay this is partially our fault, but mostly Fedora's (which is still my fault 😞 ) but I can get the `fcr:metadata` for this image
```
<http://localhost:8080/fcrepo/rest/70/80/71/78/70807178-cf92-404c-83f7-ee339e089526>
        a                        fedora:Binary , fedora:Resource , pcdm:File , ldp:NonRDFSource ;
        rdf:label                "84c5e4f3-c71e-4dc8-8abb-66ccee22717f - ServiceFile.jpeg"@en ;
        fedora:created           "2018-06-04T18:15:31.952Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:createdBy         "bypassAdmin"^^<http://www.w3.org/2001/XMLSchema#string> ;
        fedora:hasFixityService  <http://localhost:8080/fcrepo/rest/70/80/71/78/70807178-cf92-404c-83f7-ee339e089526/fcr:fixity> ;
        fedora:hasParent         <http://localhost:8080/fcrepo/rest/> ;
        fedora:lastModified      "2018-06-04T18:15:31.952Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModifiedBy    "bypassAdmin"^^<http://www.w3.org/2001/XMLSchema#string> ;
        fedora:writable          "true"^^<http://www.w3.org/2001/XMLSchema#boolean> ;
        pcdm:fileOf              <http://localhost:8000/node/2?_format=jsonld> ;
        dcterms:title            "84c5e4f3-c71e-4dc8-8abb-66ccee22717f - ServiceFile.jpeg"@en ;
        schema:additionalType    <http://localhost:8000/taxonomy/term/3?_format=jsonld> ;
        schema:author            <http://localhost:8000/user/1?_format=jsonld> ;
        schema:dateCreated       "2018-06-04T18:15:30+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        schema:dateModified      "2018-06-04T18:15:30+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        ebucore:hasMimeType      "image/jpeg"@en ;
        ebucore:height           "1333"^^<http://www.w3.org/2001/XMLSchema#int> ;
        ebucore:width            "800"^^<http://www.w3.org/2001/XMLSchema#int> ;
        iana:describedby         <http://localhost:8080/fcrepo/rest/70/80/71/78/70807178-cf92-404c-83f7-ee339e089526/fcr:metadata> ;
        premis:hasMessageDigest  <urn:sha1:244af64287d033df277668f1141c259ce01edbd7> ;
        premis:hasSize           "166608"^^<http://www.w3.org/2001/XMLSchema#long> .
```
It's just when getting the actual binary it fails.

whikloj:
Also, we need to tag a new version of Syn and start pulling that into claw-playbook because
```
> curl -i -H"Content-type: image/jpeg" -d "@louis_riel_tn.jpg" -XPUT http://localhost:8080/fcrepo/rest/jared_test
HTTP/1.1 201 Created
Server: Apache-Coyote/1.1
ETag: "1b2c15574ebe85f47a5581350b249edd34ead98f"
Last-Modified: Mon, 04 Jun 2018 19:39:01 GMT
Link: <http://localhost:8080/fcrepo/rest/jared_test/fcr:metadata>; rel="describedby"; anchor="http://localhost:8080/fcrepo/rest/jared_test"
Location: http://localhost:8080/fcrepo/rest/jared_test
Content-Type: text/plain
Content-Length: 44
Date: Mon, 04 Jun 2018 19:39:01 GMT

http://localhost:8080/fcrepo/rest/jared_test
```
I shouldn't be able to do that without authentication.

whikloj:
Ok, so other than sending language with the mimetype and the subsequent problems in Fedora this seems to work. 
@seth-shaw-unlv are you good and if so do you want to press the button? Then we cut a new version and restart Travis on islandora_image and see if it comes up green.

seth-shaw-unlv:
Let me spin up a fresh test, just to be sure, and then I will.

PR #87, opened 2018-06-07
BURN PHP 5.6 - Making our search api dependency match that of search_api_solr

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/841

* Other Relevant Links: https://github.com/Islandora-Devops/claw-playbook/pull/66

Makes composer dependencies on the `search_api` module line up with what we're installing in claw-playbook.

# What's new?
Different version for `search_api`

# How should this be tested?

Try installing with claw-playbook and you'll get past the error reported in https://github.com/Islandora-CLAW/CLAW/issues/841... and on to the next one I'm sure :raised_hands: 

# Interested parties
@DigitLib  @Islandora-CLAW/committers


dannylamb:
Kind of a hard one to test since it needs to be merged first before we can install with it.   Really I'll be doing that in https://github.com/Islandora-Devops/claw-playbook/pull/66

dannylamb:
TFW you're waiting for Travis
image

whikloj:
TFW Deprecation notices make your tests fail.

dannylamb:
<image src="http://www.dotgif.org/wp-content/uploads/2011/02/Arnold_DrosteEffect_DGDO.gif" />

dannylamb:
Whelp, updating everything to newest requires removing php 5.6 :fire::fire::fire:

whikloj:
@dannylamb did these tests run locally for you? I said it was the deprecation notices, but I didn't actually think it was...it appears the test is actually failing...is it?

dannylamb:
So deprecation notices make tests fail.  You can tell it to expect a deprecation warning through an annotation, but that doesn't appear to be working for me at the moment.

whikloj:
Can we make the call to remove support for 5.6? Do we need to go up the chain? Or to TAG?

dannylamb:
@whikloj So here's what I've done:
- pin down `search_api` to something that won't throw deprecation warnings
- removed 5.6 from the test matrix because the `search_api_solr` we need in claw-playbook requires 7

I say we let the committers decide here.  I'll update the Coordinating Committee and if this is controversial we can assemble TAG.  Considering we've talked about this lots already, I don't think it'll be a big deal but would prefer to do this as transparantly as possible.

DigitLib:
@dannylamb  Checked on both distros VMs: installed PHP is 7.0.30

whikloj:
It's loading PHPUnit 6 at least
```
 - Installing phpunit/phpunit (6.5.8): Downloading (connecting...)Downloading (100%)
```

composer.json
```diff
@@ -30,7 +30,7 @@
     "drupal/permissions_by_term" : "^1.51"
   },
   "require-dev": {
-    "phpunit/phpunit": "^4.8",
+    "phpunit/phpunit": "^6",
```
whikloj: Are we reverting this too or leaving it alone?

dannylamb: I'd keep it at 6, 4.8 is beyond EOL

dannylamb:
OK, this will need to get merged before I can tackle https://github.com/Islandora-Devops/claw-playbook/pull/66

dannylamb:
7.2 is green :raised_hands:

whikloj:
claw-burn

dannylamb:
We just keep milking that gif

PR #90, opened 2018-06-25
make csv path relative

seth-shaw-unlv:
**GitHub Issue**: Islandora-CLAW/CLAW/issues/861

Changes the migration config for tags.csv to use a relative path.

# What's new?

A relative path for tags.csv

# How should this be tested?

+ setup a new Drupal site; preferably one that doesn't use '/var/www/html/drupal/' which is part of the old hard-coded path. _Note: you don't need the whole CLAW to test this, just a vanilla Drupal works for the purposes of this test. I used a stock drupalvm site._
+ import the islandora core modules and feature
+ Run the islandora_tags migration
+ migration should run successfully with the new relative path.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/modules/islandora_core_feature/config/install/migrate_plus.migration.islandora_tags.yml
+++ b/modules/islandora_core_feature/config/install/migrate_plus.migration.islandora_tags.yml
@@ -13,7 +13,7 @@ migration_group: islandora
 label: 'Tags migration from CSV'
 source:
   plugin: csv
-  path: /var/www/html/drupal/web/modules/contrib/islandora/migrate/tags.csv
+  path: modules/contrib/islandora/migrate/tags.csv
   header_row_count: 1
   keys:
     - external_uri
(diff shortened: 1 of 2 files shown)


dannylamb:
Hey, nice!  I wasn't sure if you could do that or not.  All the examples I had when running it either used a full hardcoded path, or did silly things like move the csv to the public filesystem as part of hook_install and requested it localhost.

dannylamb:
@seth-shaw-unlv Can you do the same here? https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/modules/islandora_demo_feature/config/install/migrate_plus.migration.islandora_demo_tags.yml#L16

seth-shaw-unlv:
Sure, probably later this afternoon.

seth-shaw-unlv:
@dannylamb done.

PR #92, opened 2018-07-13
Flysystem

dannylamb:
**GitHub Issue**: Makes https://github.com/Islandora-CLAW/CLAW/issues/769 (and much more) possible

Provides a Flysystem adapter for Fedora, updates the REST api to be Flysystem aware, adds some new Condition plugins for use in context, and exports the config to glue it all together.

# What's new?
A in-depth description of the changes made by this PR. Technical details and
 possible side effects.

* Flysystem adapter for Fedora, wtih Fedora being the default upload location for files from the UI
* REST API for updating Media with binaries now accepts a `Content-Location` header, whose value is a Drupal stream wrapper, so the client can dictate where the file gets deposited.
* FileUsesFilesystem and MediaUsesFilesystem conditions, that let you filter Media and Files based on what filesystem they reside on.  You can use it to isolate Files/Media in Fedora vs. external content.
* Changed the format of AS2 events that get pumped out.  They were providing extra info that can just get grabbed using HEAD requests, so I've also updated Alpaca accordingly.

# How should this be tested?

Testing instructions to come.  I'll have to whip up a script to use until things get merged and we then we can work on claw-playbook.

# Additional Notes:
Needs tests and all. But it's time to get this up and out, just so folks know it's there and real.

# Interested parties


composer.json
```diff
@@ -27,7 +27,10 @@
     "drupal/migrate_plus" : "4.0-beta3",
     "drupal/migrate_tools" : "4.0-beta3",
     "drupal/migrate_source_csv" : "^2.1",
-    "drupal/permissions_by_term" : "^1.51"
+    "drupal/permissions_by_term" : "^1.51",
+    "drupal/token" : "^1.3",
+    "drupal/flysystem" : "^1.0",
+    "islandora/chullo" : "dev-master"
```
whikloj: We depend on `islandora/chullo` here, but it is not added to the dependencies that are installed in `islandora.info.yml` below?

modules/islandora_core_feature/config/install/field.storage.media.field_mime_type.yml
```diff
@@ -17,7 +17,7 @@ settings:
 module: core
 locked: false
 cardinality: 1
-translatable: false
+translatable: true
```
whikloj: Didn't you disable translation on mime-type to avoid mime-type having a language code added, or is that resolved because Alpaca is not `PUT`ing these objects into Fedora anymore rather Drupal is?

dannylamb: Hrm... JSONLD is not exporting the lang, and Fedora derivatives are working as expected.  I'm going to verify what's in active config for me vs. the feature.

dannylamb: I dunno what's going on.  My site has content translation turned on, and only titles and descriptions are translatable.  Mimetypes definitely do not have a box checked...

Either that config doesn't mean what I think it means, or something's totally out of whack.  It may be best for me to blow up what I've got and rebuild to make sure this is ok.

src/Flysystem/Adapter/FedoraAdapter.php
```diff
@@ -0,0 +1,325 @@
...
+    return $meta;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function listContents($directory = '', $recursive = FALSE) {
+    // Strip leading and trailing whitespace and /'s.
+    $normalized = trim($directory);
+    $normalized = trim($normalized, '/');
```
whikloj: Minor suggestion but I think you can just do 
```php
$normalized = trim($directory, ' \t\n\r\0\x0B/');
```
to do the [normal whitespace trimming](http://php.net/manual/en/function.trim.php) and the slash at the same time.

dannylamb: noice

whikloj:
Argh I was just trying to get these Contexts to leave the core block placement alone. I'll pull in these new changes. It is rather confusing.

dannylamb:
@whikloj The core plugin context stuff is super wacky.  Every time I go there I have to re-learn all of it because my brain forgets it as a self-defense mechanism.

I have to add the check for the empty form and return true on a handful of other conditions to completely unlock core block placement.   But this is the big one.

dannylamb:
@whikloj That should do it.  I just used core block placement to do something stupid like this
screenshot from 2018-08-09 14-35-31

modules/islandora_core_feature/config/install/field.storage.media.field_mime_type.yml
dannylamb: FYI I guess that's just how this works. Full rebuild seems to be ok and all the config gets imported.  Still gonna keep my :eye: on it, though.

whikloj:
Ok now it seems that the media is not being synced to Fedora but the files are, is that correct?

whikloj:
Oh forget it, I forgot to run the external bash script. Rebuilding now.

src/Plugin/Condition/ContentEntityType.php
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    if (count($this->configuration['types']) > 1) {
+      $types = $this->configuration['types'];
+      $last = array_pop($types);
+      $types = implode(', ', $filesystems);
```
whikloj: Did I mention I love PHPStorm. `$filesystems` is not defined, probably needs to be `$types`

whikloj:
Also lines [67-81](https://github.com/Islandora-CLAW/islandora/pull/92/files#diff-366174d545756e3d90e9c69aba3709ddR67) of the islandora.schema.yml need to be removed. Is causing a WSOD on the homepage for me.

dannylamb:
Nice.  I didn't do a full rebuild after fixing core block placement.  Let me take care of that.

whikloj:
Ok, this works but the [claw-playbook installation instructions](https://github.com/Islandora-Devops/claw-playbook/pull/72) require you to do the following after your run the bash script.

1. `cd /var/www/html/drupal`
1. `drupal module:uninstall islandora_demo_feature islandora_image islandora_core_feature islandora`
1. `drupal module:install islandora islandora_core_feature islandora_image islandora_demo_feature`
1. `drupal migrate:rollback --source-base_path=/var/www/html/drupal islandora_tags`
1. `drupal migrate:rollback --source-base_path=/var/www/html/drupal islandora_demo_tags`
1.  ```
    drupal migrate:execute --source-base_path=/var/www/html/drupal --site-url=http://localhost:8000 --db-type=mysql --db-host=localhost --db-name=drupal8 --db-user=drupal8 --db-pass=islandora --db-port=3306 

    Database Prefix []:
    > 

    Migration Id [all]:
    > 
    Migration Id to exclude (press <return> to stop adding migrations to exclude) []:
    > 
    Processing Migration "islandora_demo_tags"
    Migration "islandora_demo_tags" was imported correctly
    Processing Migration "islandora_tags"
    Migration "islandora_tags" was imported correctly
    ```
1. `drupal features:import islandora_core_feature`
1. `drupal features:import islandora_demo_feature`

then you're ready 👌

PR #106, opened 2018-12-06
Make derivative Drupal media type configurable

whikloj:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/986

To allow for Image and Video events to share code we (I) tried to use the source media's bundle. But this causes problems when your source media is one type but the derivative should be another.

This PR adds a configuration option (Derivative media type) to the Action to set the destination media type which is used in the file_upload_uri when putting a new File into Drupal.

# What's new?

* Does this change require documentation to be updated? no
* Does this change add any new dependencies? no
 (ie. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? no

# How should this be tested?

Before the PR, adding a Tiff as a File and you would end up with no derivatives and a lot of errors in the log.

After you should get your correct derivatives back in Drupal.

# Interested parties
@Islandora-CLAW/committers


Natkeeran:
@whikloj Looks good.  Travis is complaining.

dannylamb:
@whikloj @Natkeeran I can also confirm that this is working as intended.  I uploaded a Tiff as a File and got Image derivatives.

PR #107, opened 2018-12-11
Allow JSON-LD @Type to be configurable based on a specified Content Entity field

seth-shaw-unlv:
**GitHub Issue**: [Issue 889](https://github.com/Islandora-CLAW/CLAW/issues/889)

Adds a new Condition Plugin allowing users to specify which bundles apply. Adds a new Reaction Plugin to alter the JSON-LD's `@type` attribute based on the predicate values in the configured field.

# What's new?

* Added
    - EntityBundle Condition Plugin
    - JsonldTypeAlterReaction Reaction Plugin
* Does this change require documentation to be updated? maybe?
* Does this change add any new dependencies? no.
 (ie. Regeneration activity, etc.)? Only if you want it.
* Could this change impact execution of existing code? no.

# How should this be tested?

- Apply this PR to the islandora module.
- Clear cache (otherwise the new condition and reaction plugins won't show up in your lists later)
- Either:
    - Create a field on a content type or taxonomy where a user can enter or select a type predicate,
    - OR Checkout the [controlled_access_terms issue-889 branch](https://github.com/Islandora-CLAW/controlled_access_terms/tree/issue-889) and feature import the changes. (Creates a `field_type` field for Corporate Bodies with some default schema predicates.)
- Create a new Context using, at least, the JsonldTypeAlterReaction with the new field you created. If you are using the setup provided by controlled_access_terms issue-889, then you can import the config below:
```
uuid: c23f9c1c-1dbb-4958-a2e8-639fc4c291a6
langcode: en
status: true
dependencies:
  module:
    - islandora
name: alter_json_ld_type
label: 'Alter JSON-LD Type'
group: Islandora
description: 'Adds additional @type to JSON-LD based on configured field'
requireAllConditions: false
disabled: false
conditions:
  entity_bundle:
    id: entity_bundle
    bundles:
      corporate_body: corporate_body
    negate: 0
    uuid: ddb49e9e-0d59-4490-93bb-22a2ceaa6c1d
    context_mapping:
      node: '@node.node_route_context:node'
      taxonomy_term: '@islandora.taxonomy_term_route_context_provider:taxonomy_term'
reactions:
  alter_jsonld_type:
    id: alter_jsonld_type
    source_field: field_type
    saved: false
weight: 0
```
- Create a new node/taxonomy_term that uses the configured field.
- Viewing the node/taxonomy_term's JSON-LD will show the additional types.
- Searching for it in Blazegraph or Fedora will also display the additional types.

# Interested parties
@Islandora-CLAW/committers (esp @dannylamb and @patdunlavey)

### DIFF
--- a/src/Plugin/Condition/EntityBundle.php
+++ b/src/Plugin/Condition/EntityBundle.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Drupal\islandora\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a 'Entity Bundle' condition.
+ *
+ * @Condition(
+ *   id = "entity_bundle",
+ *   label = @Translation("Entity Bundle"),
+ *   context = {
+ *     "node" = @ContextDefinition("entity:node", label = @Translation("Node")),
+ *     "taxonomy_term" = @ContextDefinition("entity:taxonomy_term", label = @Translation("Term"))
+ *   }
+ * )
+ */
+class EntityBundle extends ConditionPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $options = [];
+    foreach (['node', 'taxonomy_term'] as $content_entity) {
+      $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($content_entity);
+      foreach ($bundles as $bundle => $bundle_properties) {
+        $options[$bundle] = $this->t('@bundle (@type)', [
+          '@bundle' => $bundle_properties['label'],
+          '@type' => $content_entity,
+        ]);
+      }
+    }
+
+    $form['bundles'] = [
+      '#title' => $this->t('Bundles'),
+      '#type' => 'checkboxes',
+      '#options' => $options,
+      '#default_value' => $this->configuration['bundles'],
+    ];
+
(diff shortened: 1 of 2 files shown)


dannylamb:
This is nice, simple, and flexible.  I added a "manual override" field to subjects and set one to `skos:Concept`.

My Context:
screenshot from 2018-12-12 16-06-31

JSONLD Before:
screenshot from 2018-12-12 16-04-16

JSONLD After:
screenshot from 2018-12-12 16-06-00

Let me see if there's a test that can easily be copy/pasta'd for the other jsonld alter we're doing.  It would be nice to have some code coverage on these plugins.

seth-shaw-unlv:
So, @dannylamb, I was just thinking. During the call you asked if this PR was using Taxonomy terms with URIs, which this doesn't. But should it? It seems that using Taxonomy would be more consistent with the paradigms we've set for controlling what URIs we want to use for things. I could fairly easily refactor this to use them instead of a configured field. 

What do you think?

dannylamb:
@seth-shaw-unlv That's what I was originally thinking we'd do, and wire that up with `field_model` and `field_media_use` by default.  But I like what this does too and I see its benefit.  How about we keep what you've got and add a check for entity references with a `field_external_uri`, and use that instead if it exists?

seth-shaw-unlv:
@dannylamb , we could do that. I suppose we would check if `array_key_exists('target_id')` and, if so, load the target entity to check and see if it contains a `field_external_uri`. Does that sound right? Would we presume that the `field_external_uri` is fully qualified, or should we normalize it, too?

PR #113, opened 2019-02-04
Work on https://github.com/Islandora-CLAW/CLAW/issues/1014.

mjordan:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1014

Fixes URIs persisted to Gemini like `http://localhost:8080/fcrepo/rest//islandora_2_MODS.xml` so that they don't contain double forward slashes (`//`).

# What's new?

Within the `src/Plugin/Action/EmitFileEvent.php` class, checks to see if the incoming URI for the file contains a redundant `/`, and if so, removes it.

# How should this be tested?

Since this error manifested itself in a migration from a 7.x instance, you will need one of those to test this fix. 

1. Install the Migrate 7.x CLAW module (in /var/www/html/drupal, run `composer require islandora/migrate_7x_claw`)
1. Configure your migration as described in https://github.com/Islandora-Devops/migrate_7x_claw.
1. Enable the migration modules by running `drush en islandora_migrate_7x_claw_feature`.
1. Migrate some objects from an Islandora 7.x instance.
1. Fire up the MySQL client on the CLAW  VM and in the gemini database, issue the query `select * from Gemini\G`. You should see values in the `fedora_uri` fields that contain double forward slashes (`//`), e.g. `http://localhost:8080/fcrepo/rest//islandora_2_MODS.xml`.
1. Roll back your migration by running `drush -y migrate:rollback --group islandora_7x`.
1. Within /var/www/html/drupal/web/modules/contrib/islandora, run `sudo git reset --hard` to reset permissions on files changed by Ansible (?).
1. Check out the `mjordan:issue-1014` branch.
1. Rerun your migration by running `drush -y mim --userid=1 --group islandora_7x`.
1. Check your gemini database again. No `fedora_uri` values should contain double slashes (`//`), and all URIs should look otherwise correct.

# Interested parties
@dannylamb, other @Islandora-CLAW/committers

### DIFF
--- a/src/Plugin/Action/EmitFileEvent.php
+++ b/src/Plugin/Action/EmitFileEvent.php
@@ -104,7 +104,8 @@ class EmitFileEvent extends EmitEvent {
 
     $data = parent::generateData($entity);
     if (isset($flysystem_config[$scheme]) && $flysystem_config[$scheme]['driver'] == 'fedora') {
-      // $uri for files may contain 'fedora:///' so we need to replace the three / with two.
+      // $uri for files may contain 'fedora:///' so we need to replace
+      // the three / with two.
       if (strpos($uri, 'fedora:///') !== FALSE) {
         $uri = str_replace('fedora:///', 'fedora://', $uri);
       }
--- a/src/Plugin/Action/EmitFileEvent.php
+++ b/src/Plugin/Action/EmitFileEvent.php
@@ -104,10 +104,10 @@ class EmitFileEvent extends EmitEvent {
 
     $data = parent::generateData($entity);
     if (isset($flysystem_config[$scheme]) && $flysystem_config[$scheme]['driver'] == 'fedora') {
-      // $uri for files may contain 'fedora:///' so we need to replace
+      // Fdora $uri for files may contain ':///' so we need to replace
       // the three / with two.
-      if (strpos($uri, 'fedora:///') !== FALSE) {
-        $uri = str_replace('fedora:///', 'fedora://', $uri);
+      if (strpos($uri, $scheme . ':///') !== FALSE) {
+        $uri = str_replace($scheme . ':///', $scheme . '://', $uri);
       }
       $data['fedora_uri'] = str_replace("$scheme://", $flysystem_config[$scheme]['config']['root'], $uri);
     }
--- a/src/Plugin/Action/EmitFileEvent.php
+++ b/src/Plugin/Action/EmitFileEvent.php
@@ -104,6 +104,10 @@ class EmitFileEvent extends EmitEvent {
 
     $data = parent::generateData($entity);
     if (isset($flysystem_config[$scheme]) && $flysystem_config[$scheme]['driver'] == 'fedora') {
+      // $uri for files may contain 'fedora:///' so we need to replace the three / with two.
+      if (strpos($uri, 'fedora:///') !== FALSE) {
+        $uri = str_replace('fedora:///', 'fedora://', $uri);
+      }
       $data['fedora_uri'] = str_replace("$scheme://", $flysystem_config[$scheme]['config']['root'], $uri);
     }
     return $data;


mjordan:
"The command "phpcs --standard=Drupal --ignore=*.md --extensions=php,module,inc,install,test,profile,theme,css,info $TRAVIS_BUILD_DIR" exited with 1". I will investigate.

src/Plugin/Action/EmitFileEvent.php
```diff
@@ -104,6 +104,10 @@ protected function generateData(EntityInterface $entity) {
 
     $data = parent::generateData($entity);
     if (isset($flysystem_config[$scheme]) && $flysystem_config[$scheme]['driver'] == 'fedora') {
+      // $uri for files may contain 'fedora:///' so we need to replace the three / with two.
```
seth-shaw-unlv: This is the offending line for phpcs (comment lines can't go beyond the 80 character line). Just break it into two lines and you will be fine.

mjordan: @seth-shaw-unlv sweet, thanks!

mjordan: That did it, thanks again @seth-shaw-unlv.

src/Plugin/Action/EmitFileEvent.php
```diff
@@ -104,6 +104,11 @@ protected function generateData(EntityInterface $entity) {
 
     $data = parent::generateData($entity);
     if (isset($flysystem_config[$scheme]) && $flysystem_config[$scheme]['driver'] == 'fedora') {
+      // $uri for files may contain 'fedora:///' so we need to replace
+      // the three / with two.
+      if (strpos($uri, 'fedora:///') !== FALSE) {
```
dannylamb: You should use `"$scheme"` instead of `fedora` in these operations.  This works out now because `fedora` is the default, but you can potentially define a different scheme name in `settings.php`.  We should stick to what's defined there.

mjordan: OK, thanks, will change it.

mjordan: But I will point out that it's used within an `if` block that uses "$flysystem_config[$scheme]['driver'] == 'fedora'".

dannylamb: There's a subtle difference between the flysystem scheme and the flysystem driver.  The scheme is what comes before :// in the uri, but the driver is the id of the plugin used in flysystem.  They're both fedora right now, but conceivably you could have 'fedoraA' and 'fedoraB' schemes that both use the 'fedora' driver.

mjordan: OK, got it.

dannylamb:
@mjordan Re: tests from the last CLAW call, looks like when I refactored how we generate the events, I cleared out some tests because now one method pretty much handles everything.  Before it was unneccessarily broken apart.  If you look back in history, though, there's examples of triggering events and sniffing the messages that get published to the queue.  https://github.com/Islandora-CLAW/islandora/commit/a1987aecb747b0b297176fc4d34afdc22cbddb4d#diff-d936ab04878146b04d45ae403558ff54  That's probably the best starting point to test the file events and guarantee no `///`s get introduced.

mjordan:
@dannylamb OK, just pushed up the code that uses $scheme.

WRT tests, before I start figuring out how to write a test for this, I thought I'd figure out how to run tests locally. In `vagrant@claw:/var/www/html/drupal/web`, I ran  `php core/scripts/run-tests.sh islandora`, but they all fail:

```
Test summary
------------

Drupal\Tests\islandora\Functional\AddMediaToNodeTest           0 passes             1 exceptions             
FATAL Drupal\Tests\islandora\Functional\AddMediaToNodeTest: test runner returned a non-zero error code (2).
Drupal\Tests\islandora\Functional\AddMediaToNodeTest           0 passes   1 fails                            
Drupal\Tests\islandora\Functional\ContentEntityTypeTest        0 passes             1 exceptions             
FATAL Drupal\Tests\islandora\Functional\ContentEntityTypeTest: test runner returned a non-zero error code (2).
Drupal\Tests\islandora\Functional\ContentEntityTypeTest        0 passes   1 fails                         
```
etc. How do you run tests locally?

dannylamb:
@mjordan I copy what travis does, but run it as the apache user:

```bash
vagrant@claw:/var/www/html/drupal/web$ sudo -u www-data php core/scripts/run-tests.sh --suppress-deprecations --url http://127.0.0.1:8000 --verbose --php `which php` --module "islandora"
```

mjordan:
@dannylamb thanks, running tests with that worked. We should include that in the developer docs.

Over the weekend I'll see if I can write a test for this PR.

mjordan:
@dannylamb I see you're merging a bunch of stuff. If you want to merge this without tests, go ahead, I can add those in a separate PR. Your call.

dannylamb:
@mjordan Give the tests a go and see how it turns out.  You should be able to copy/paste your way through.  And don't be shy to reach out. Functional tests are tricky beasts.

mjordan:
@dannylamb I'm not going to have time in the next while to dig into the tests. I know we want as many tests as possible, but this is a fix, not a new feature, so I'm wondering if we should merge this as is so it doesn't get stale.

dannylamb:
@mjordan Hit the update branch button and we'll go

mjordan:
Button hit. Thanks @dannylamb.

PR #116, opened 2019-02-22
Make JWT Expiry configurable - Issue 1030

seth-shaw-unlv:
**GitHub Issue**: /Islandora-CLAW/CLAW/issues/1030

See also the IRC conversations linked from the issue.

This PR amends the Islandora core setting form to include the option of setting a JWT expiration time (default is currently hard-coded as '+2 hour') and updates the JwtEventSubscription to use the setting (if available, the default '+2 hour' if not).

# What's new?

* Changes JwtEventSubscriber feature to such that the expiration is pulled from configuration.
* Added JWT Setting to Islandora core settings form.

# How should this be tested?

- Add the debug setting to Chullo FedoraApi class. Add `$options['debug'] = TRUE;` [about here](https://github.com/Islandora-CLAW/chullo/blob/master/src/FedoraApi.php#L76).
- Start a migration of any sort that pushes to Fedora (you can limit it to just one item for debugging purposes).
- You will start to see authorization tokens appear `Authorization: Bearer ...`. Decode the JWT. The [JWT homepage](https://jwt.io/) has a decoding/debugger tool that makes this easy; just copy paste in there and it will show you the decoded content. The exipry is given as an epoc int, but mouse over the value to see the human-readable form. By default it should be about two hours from when the migration was started.
- Apply the PR
- Adjust the exipry setting on the Islandora core settings form. (E.g. '+2 days'.)
- Start another migration and check the authorization JWT again. The expiry should now reflect the update expiry setting from your form.

# Interested parties
@dannylamb or any other @Islandora-CLAW/committers

### DIFF
--- a/src/EventSubscriber/JwtEventSubscriber.php
+++ b/src/EventSubscriber/JwtEventSubscriber.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\islandora\EventSubscriber;
 
+use Drupal\islandora\Form\IslandoraSettingsForm;
 use Drupal\jwt\Authentication\Event\JwtAuthValidateEvent;
 use Drupal\jwt\Authentication\Event\JwtAuthValidEvent;
 use Drupal\jwt\Authentication\Event\JwtAuthGenerateEvent;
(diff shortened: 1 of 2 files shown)


seth-shaw-unlv:
Even though #119 made it so that JWT expiry was tied to event generation rather than the migration start time, this PR is still valuable until we come up with something better as per [issue 1036](https://github.com/Islandora-CLAW/CLAW/issues/1036).

A bug, that requires its own ticket, is that running media migrations while derivatives are enabled creates database locking errors because the migration and derivatives processes are trying to take the same media ids. The work-around is to disable derivatives contexts until the migration completes and then trigger them manually after the fact. This means I need to either emit a large number of events all at once (giving them essentially the same expiry) or batch them. 

I can process service images and thumbnails for a little more than 3000 tiffs in that two-hour window, resulting in batches of 3000 every 2 hours (or 1500 every hour, etc.). For a collection of 40k+ tiffs, that is 14 batches started every 2 hours for 28 hours. Until we can start derivatives via cron, that isn't a nice prospect for a repository manager.

OR I can (using this PR) increase the expiry setting for as long as I think it will take to do them all (plus a buffer), start them all, and then reset it back to the original 2 hour expiry.

dannylamb:
@seth-shaw-unlv Hit the update button and I'll merge this :bowing_man:

seth-shaw-unlv:
@dannylamb green means go.

PR #118, opened 2019-02-26
Flysystem Adapter JWT Every Transaction: Issue-1030

seth-shaw-unlv:
**GitHub Issue**: Islandora-CLAW/CLAW/issues/1030

Related (but completely independent) PR: Islandora-CLAW/islandora/pull/116 for increasing JWT expiry times.

Shifts the instantiation of the Chullo/Guzzle client so that a new JWT token will be issued with each request rather than reusing it over the course of several actions.

# What's new?

* Moved code associated with creating a Chullo client from Flysystem/Fedora.php to Flysystem/Adapter/FedoraAdapter.php.
* Add a getFedora function to FedoraAdapter which generates a new Chullo instance (with associated JWT) and modified previous calls to a stored instance to use this method instead.

# How should this be tested?

* Add the debug setting to Chullo FedoraApi class. Add `$options['debug'] = TRUE;` [about here](https://github.com/Islandora-CLAW/chullo/blob/master/src/FedoraApi.php#L76).
* Start a migration of any sort that pushes a file to Fedora (you can limit it to just two-three items for testing purposes).
* You will start to see authorization tokens appear `Authorization: Bearer ...`. Decode the JWT. The [JWT homepage](https://jwt.io/) has a decoding/debugger tool that makes this easy; just copy paste in there and it will show you the decoded content. All of the tokens should be the same.
* Apply the PR
* Run another migration (again, you can limit to just two or three for testing). Each token should be unique.

# Additional Notes:

I'm not completely convinced this is the way to structure everything since Flysystem/Fedora simply hands off the whole configuration array to the FedoraAdapter; but it works. I was tempted to try simply combining the two classes so that it implements all three interfaces which is closer to how the Flysystem Adapters appear to work, but this is the structure all the Drupal implementations of Flysystem Adapters use, so there we go. Give it a good hard look.

# Interested parties
@dannylamb @whikloj @jonathangreen

### DIFF
--- a/src/Flysystem/Adapter/FedoraAdapter.php
+++ b/src/Flysystem/Adapter/FedoraAdapter.php
@@ -2,13 +2,17 @@
 
 namespace Drupal\islandora\Flysystem\Adapter;
 
-use Islandora\Chullo\IFedoraApi;
+use Drupal\jwt\Authentication\Provider\JwtAuth;
+use Islandora\Chullo\FedoraApi;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Adapter\Polyfill\NotSupportingVisibilityTrait;
 use League\Flysystem\Adapter\Polyfill\StreamedCopyTrait;
 use League\Flysystem\Config;
+use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Client;
 use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
 use GuzzleHttp\Psr7\StreamWrapper;
 use Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesserInterface;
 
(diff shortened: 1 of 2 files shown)


whikloj:
I don't know much about Flysystem, but I am wondering if changing from never issuing a new JWT to always issuing a new JWT might cause some performance issues? I don't really know what the hit is for generating a JWT. 

I'm thinking of what @jonathangreen was suggesting to extend the expiry and wondering if it would be more preformant to test the current JWT's expiration and recreate only if it is expired or within a small window of being expired?

So I guess I am wondering is it easier to resolve the JWT to test the expiration or just create a new one?

whikloj:
Annnnnd I just saw [this](https://github.com/Islandora-CLAW/CLAW/issues/1030#issuecomment-467656794), so you can disregard my concerns if you all have done some investigation.

dannylamb:
I'm sad we have to couple some of our Drupal specifics in the Adapter itself, which I was attempting to keep fairly isolated.  But oh well, cuz people need to migrate for longer than 2 hours.

The code itself looks fine, you're just shuffling that function and the data you need down another layer.  You'll need to touch up a kernel test, mostly just b/c the parameters have changed so you need to mock things differently.  Also, I'm pretty sure CodeSniffer's gonna complain, but that's nothing `phpcbd` can't take care of.

Let me know if you have any questions about the kernel test.

seth-shaw-unlv:
@whikloj, it was my understanding of @jonathangreen's comment that we _shouldn't_ be reusing the JWT until it expires (thus treating it as a password rather than as a one-time use token). 

I was also thinking about the potential performance impact, but considering this bit of code is focusing on moving binaries around, I'm inclined to think that the performance hit of regenerating a token (which, really, is pretty small) each time is a drop in the bucket compared to the file transfer times involved. I highly doubt anyone would notice the difference and I am remiss to complicate the code for premature optimization until we have evidence it is problematic.

I should be able to get to those tests today.

seth-shaw-unlv:
Superseded by #119

PR #119, opened 2019-02-27
New JWT for each Flysystem request: Issue-1030

seth-shaw-unlv:
**GitHub Issue**: [Islandora-CLAW/CLAW/issues/1030](https://github.com/Islandora-CLAW/CLAW/issues/1030)

Supersedes #118 (Thanks to @whikloj for [the idea for this new approach](https://logs.woh.ca/islandora/2019-02-27#501593;)!)

Related (but completely independent) PR: #116 for increasing JWT expiry times.

Modifies the Guzzle client handler to generate a new JWT token for every request.

# What's new?
* Replaced addHeader function with addJWT.
* Modified the create function to use the new addJWT function (which now generates the token) in the HandlerStack instead of generating the token and passing it through addHeader.

# How should this be tested?
* Add the debug setting to Chullo FedoraApi class. Add `$options['debug'] = TRUE;` [about here](https://github.com/Islandora-CLAW/chullo/blob/master/src/FedoraApi.php#L76).
* Start a migration of any sort that pushes a file to Fedora (you can limit it to just two-three items for testing purposes).
* You will start to see authorization tokens appear `Authorization: Bearer ...`. Decode the JWT. The [JWT homepage](https://jwt.io/) has a decoding/debugger tool that makes this easy; just copy paste in there and it will show you the decoded content. All of the tokens should be the same.
* Apply the PR
* Run another migration (again, you can limit to just two or three for testing). Each token should be unique.

# Additional Notes:
This new approach is a super simple solution compared to the rewrite I had attempted. The last version made the kernel tests nearly impossible while this version required no updates to the tests.

# Interested parties
@dannylamb @whikloj @jonathangreen

### DIFF
--- a/src/Flysystem/Fedora.php
+++ b/src/Flysystem/Fedora.php
@@ -7,6 +7,7 @@ use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\flysystem\Plugin\FlysystemPluginInterface;
 use Drupal\flysystem\Plugin\FlysystemUrlTrait;
 use Drupal\islandora\Flysystem\Adapter\FedoraAdapter;
+use Drupal\jwt\Authentication\Provider\JwtAuth;
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Client;
 use Islandora\Chullo\IFedoraApi;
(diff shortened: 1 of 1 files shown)


seth-shaw-unlv:
Note: the testing instructions suggest adding the debug option to FedoraApi.php to view the JWT sent with each Fedora request. This works if you run migrations from Drush; HOWEVER, this breaks file uploads from the web UI (although I don't know why) so it should only be used for testing this PR and/or #116.

whikloj:
Nice work @seth-shaw-unlv 👏

dannylamb:
Fantastic @seth-shaw-unlv and @whikloj :bowing_man:

seth-shaw-unlv:
@dannylamb green means go.

dannylamb:
The CLAW patrol is on a roll:

seth-shaw-unlv:
I get Rubble.

PR #124, opened 2019-04-23
Display Fedora URI in Drupal

whikloj:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/881

# What does this Pull Request do?

Creates a pseudo field to display the linked URI from the Gemini database on the Drupal resource.

# What's new?

Create two new services, one is a GeminiClient (with factory) and one does the lookup from an entity.

Alters the Islandora Core Settings to allow for a Gemini URI (with some simple validation) and to select which node, media or taxononomy_terms can have the field displayed.

Uses two hooks in `islandora.module` to define the new fields and build them.

* Does this change require documentation to be updated? Maybe
* Does this change add any new dependencies? yes, change from chullo to crayfish-commons
* Does this change require any other modifications to be made to the repository
 (ie. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? no

# How should this be tested?

1. Build a claw-playbook.
1. Login and replace the existing islandora module directory with this branch.
1. Install crayfish-commons (should happen automatically once this is merged when we install the module).
    1. `cd /var/www/html/drupal`
    1. `composer require "islandora/crayfish-commons:^0.0"`
1. Go to the **Islandora Settings** and try adding various things to the Gemini URL, non-URLs should throw an error and URLs that are fake should throw an error.
1. Check off "Repository Item (node)" under "Bundles with Gemini URI Pseudo field"
1. Save the form.
1. Clear the cache (could not figure out a cache tag to resolve this)
1. Edit the field display for the Repository Item (Structure -> Content-Types -> Manage Display [for Repository Item]) and move the "Fedora URI" field where ever you want it.
1. Save the field display.
1. Create a new **Repository Item**.
1. Wait a couple seconds and then try reloading the item. If the field does not appear, try clearing the  cache again. This might be a separate task, how to trigger a cache clear once the async PUT occurs.
1. The **Fedora URI** field should appear and have a link to the associated item in Fedora.  This is stored in cache and not regenerated with each view.

# Additional Notes:
There are a couple caching issues that could make this better, but I was lost. 

1. Once you enable the field on a content or media type, it doesn't appear in the list of fields until you clear the cache. Even though when you request the field display page you get `Cache-Control: must-revalidate, no-cache, private` and `X-Drupal-Dynamic-Cache: UNCACHEABLE`.
1. Also I'm not sure how we might have the Drupal cache for a node cleared once it gets written to Fedora. Perhaps some small rest service that takes the entity type and id and clears the associated cache tags.

[...]

PR #126, opened 2019-05-02
Media Fedora Uri Pseudo-Fields

dannylamb:
**GitHub Issue**: 
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/881
- Resolves https://github.com/Islandora-CLAW/CLAW/issues/1109
- Follow up to https://github.com/Islandora-CLAW/islandora/pull/124

Adds the Fedora Uri pseudo-field to Media.

# What's new?
Looks up media in addition to other entity types in the GeminiLookup class.  This is done by issuing a HEAD request against the binary in Fedora to find it's `rel="describedby"` link header.

# How should this be tested?

- Pull down this PR
- Clear your cache
- You should now see the Fedora Uri pseudo-field on Media entities
Screenshot from 2019-05-02 12-20-03

@Islandora-CLAW/committers

### DIFF
--- a/src/GeminiLookup.php
+++ b/src/GeminiLookup.php
@@ -3,10 +3,14 @@
 namespace Drupal\islandora;
 
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\islandora\MediaSource\MediaSourceService;
 use Drupal\jwt\Authentication\Provider\JwtAuth;
+use GuzzleHttp\Psr7;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\RequestException;
 use Islandora\Crayfish\Commons\Client\GeminiClient;
 use Psr\Log\LoggerInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
  * Locates the matching Fedora URI from the Gemini database.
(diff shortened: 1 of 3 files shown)


src/GeminiLookup.php
```diff
@@ -77,24 +90,72 @@ public static function create(ContainerInterface $container) {
...
+      return NULL;
+    }
+
+    $is_media = $entity->getEntityTypeId() == 'media';
+
+    // Use the entity's uuid unless it's a media,
+    // use its file's uuid instead.
+    if ($is_media) {
+      try {
+        $file = $this->mediaSource->getSourceFile($entity);
```
whikloj: Why are we linking the media using the file's UUID?

whikloj:
@dannylamb you've got some duplication in your test setup. Might want to replace the local variables `$entity`, `$file`, `$media_source`, `$gemini_client` and `$guzzle` with private class variables so you can populate them in a function and call that from multiple locations.

dannylamb:
The ones that never change I kept as class variables, but the ones that did I just copy/pasta'd around in all the functions.  Let me declare them all and set defaults in `setUp`. That would reduce code for sure.

dannylamb:
Thx for keeping me on point @whikloj.  Refactored tests have been pushed up.  Let me know what you think.

PR #127, opened 2019-05-03
Slideshow image formatter

dannylamb:
**GitHub Issue**: An attempt to resolve https://github.com/Islandora-CLAW/CLAW/issues/1032

Adds a new Image field formatter that renders image as links to the Node's content and not the Media.  This means you can make a view that queries for all media with X Properties that belong to a node with Y Properties, etc... And have those images lead to the _node_ when clicked.  This means we can make teaser mode lead to the full content of the link when the user clicks on the thumbnail in it.  It also means we can make galleries or slideshows using views super easily.  If you're capable of contextual filters and relationships in views, you can do _a ton_ of things.

# What's new?
An image field formatter that interprets the "Content" selection for link settings as pointing to the node that's the 'field_media_of' as opposed to the media itself.

Tests...

# How should this be tested?

Make a view!

Get some content in:
- Make it for Media
- Set it to grid
- Set it to render fields, and then the `field_media_image` field using the Islandora Image formatter
- Set a filter for 'field_media_use' to 'Thumbnail Files'
- Voila, now you have a view that spits out a grid of thumbnails, each one as a link leading to their node

You can do some crazy stuff with Contextual Filters and Relationships, too.  Allowing you to filter by what's on the node that owns a media, as well as what's on the media itself.  It's sweet.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/config/schema/islandora.schema.yml
+++ b/config/schema/islandora.schema.yml
@@ -120,3 +120,14 @@ condition.plugin.content_entity_type:
       type: sequence
       sequence:
         type: string
+
+field.formatter.settings.islandora_image:
+  type: mapping
+  label: 'Image field display format settings'
+  mapping:
+    image_link:
+      type: string
+      label: 'Link image to'
+    image_style:
+      type: string
+      label: 'Image style'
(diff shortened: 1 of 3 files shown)


seth-shaw-unlv:
I will tentatively say this works great, having worked with it before @dannylamb issued the PR. I'll give it a more formal test and review next Monday unless someone beats me to it.

PR #128, opened 2019-05-07
Video frame extraction

dannylamb:
**GitHub Issue**: Part of https://github.com/Islandora-CLAW/CLAW/issues/1005

Loosens the restrictions around what types of mimetypes can be requested from Homarus, since we can totally extract images using `-ss 00:00:01.000 -frames 1`.

# How should this be tested?

Now when you edit a "Generate video derivative" action, you are allowed to enter `video/*` OR `image/jpeg` if you want to extract a frame from the video.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/config/schema/islandora.schema.yml
+++ b/config/schema/islandora.schema.yml
@@ -120,3 +120,14 @@ condition.plugin.content_entity_type:
       type: sequence
       sequence:
         type: string
+
+field.formatter.settings.islandora_image:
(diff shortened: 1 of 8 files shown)


dannylamb:
Closing, as this is unneccessary if I use the Image action and tell it to point to the homarus queue instead.  No need to butcher the video action's validate function to push things through.

PR #132, opened 2019-05-10
Breadcrumbs sub-module

seth-shaw-unlv:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1141

Provides an islandora_breadcrumb submodule. The breadcrumbs are created by following the field_member_of chain from the current node until it runs out of links.

# What's new?
* Added the islandora_breadcrumbs submodule.
* Does this change require documentation to be updated? We could add some....
* Could this change impact execution of existing code? No, it is completely optional and only impacts other breadcrumb settings you may have.

# How should this be tested?

* Make a collection
* Make a sub-collection belonging to the first collection (repeat as often as you would like)
* Make an item belonging to any of those collections
* Be sad that the item has no breadcrumbs.
* Apply the PR
* Enable the Islandora Breadcrumbs module (e.g. `drush en-y islandora_breadcrumbs`)
* Flush caches
* View the item
* Rejoice, for there are breadcrumbs!

# Additional Notes

We could add some configuration to it later, but this will probably be fine for now.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,8 @@ install:
   - cd web; drush --uri=127.0.0.1:8282 en -y islandora
 
 script:
-  - $SCRIPT_DIR/line_endings.sh $TRAVIS_BUILD_DIR
-  - phpcs --standard=Drupal --ignore=*.md --extensions=php,module,inc,install,test,profile,theme,css,info $TRAVIS_BUILD_DIR
-  - phpcpd --names *.module,*.inc,*.test,*.php $TRAVIS_BUILD_DIR
-  - php core/scripts/run-tests.sh --suppress-deprecations --url http://127.0.0.1:8282 --verbose --php `which php` --module "islandora"
+  - $SCRIPT_DIR/travis_scripts.sh
+  - $SCRIPT_DIR/run-tests.sh "islandora_breadcrumbs"
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install:
 
 script:
   - $SCRIPT_DIR/travis_scripts.sh
+  - $SCRIPT_DIR/run-tests.sh "islandora"
   - $SCRIPT_DIR/run-tests.sh "islandora_breadcrumbs"
 
 notifications:
--- a/modules/islandora_breadcrumbs/islandora_breadcrumbs.services.yml
+++ b/modules/islandora_breadcrumbs/islandora_breadcrumbs.services.yml
@@ -1,0 +1,5 @@
+services:
+  islandora_breadcrumbs.breadcrumb:
+    class: Drupal\islandora_breadcrumbs\IslandoraBreadcrumbBuilder
+    tags:
+      - { name: breadcrumb_builder, priority: 100 }
--- a/modules/islandora_breadcrumbs/islandora_breadcrumbs.services.yml
+++ b/modules/islandora_breadcrumbs/islandora_breadcrumbs.services.yml
@@ -1,5 +1,6 @@
 services:
   islandora_breadcrumbs.breadcrumb:
     class: Drupal\islandora_breadcrumbs\IslandoraBreadcrumbBuilder
+    arguments: ['@config.factory']
     tags:
       - { name: breadcrumb_builder, priority: 100 }
--- a/modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
+++ b/modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
@@ -29,7 +29,9 @@ class IslandoraBreadcrumbBuilder implements BreadcrumbBuilderInterface {
   public function build(RouteMatchInterface $route_match) {
     $node = $route_match->getParameter('node');
     $breadcrumb = new Breadcrumb();
-    $chain = IslandoraBreadcrumbBuilder::walkMembership($node);
+
+    $chain = [];
+    IslandoraBreadcrumbBuilder::walkMembership($node, $chain);
 
     // Don't include the current item. @TODO make configurable.
     array_pop($chain);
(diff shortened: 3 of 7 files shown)


whikloj:
I am wondering if you create a Collection A which has a member Collection B which has a Collection C, then make Collection A a member of Collection C

I feel like you would end up in a loop, probably need an escape hatch.

seth-shaw-unlv:
@whikloj, I added an escape for when we hit a repeated entity in the chain.

seth-shaw-unlv:
We could also add a configured depth limit break after the array_unshift later:
```
if (count($crumbs) >= $depth_limit) {
 return;
}
```
Once we get around to adding a config file...

seth-shaw-unlv:
Ok, I went ahead and added some basic config support. However, we don't have a config form yet or a default config file.

modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
```diff
@@ -0,0 +1,89 @@
...
+              !$parameters['node']->field_member_of->isEmpty());
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(RouteMatchInterface $route_match) {
+
+    $config = \Drupal::config('islandora.breadcrumbs');
```
whikloj: You can pass an active instance of the config factory in to this service to avoid these static calls. An example of almost exactly this from jsonld (https://github.com/Islandora-CLAW/jsonld/commit/faa8301eb7f6425d99ff416f2bb254cbd39e96f7#diff-4b4918881456938d1c1b5fb01d3ca194R64)

modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
```diff
@@ -0,0 +1,89 @@
...
+    if (!empty($config->get('depthLimit'))) {
+      $this->depthLimit = $config->get('depthLimit');
+    }
+    if (!empty($config->get('includeSelf'))) {
+      $this->includeSelf = $config->get('includeSelf');
+    }
```
whikloj: You should include a default config install schema file. That way you can assign those defaults there and then you don't need to check for empty here.

modules/islandora_breadcrumbs/islandora_breadcrumbs.services.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  islandora_breadcrumbs.breadcrumb:
+    class: Drupal\islandora_breadcrumbs\IslandoraBreadcrumbBuilder
+    tags:
```
whikloj: You'll need to add the `arguments` here to inject the `@config.factory` service into your service

modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
```diff
@@ -0,0 +1,89 @@
...
+  protected $depthLimit = 2;
+  protected $includeSelf = FALSE;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applies(RouteMatchInterface $attributes) {
+    $parameters = $attributes->getParameters()->all();
+    if (!empty($parameters['node'])) {
+      return ($parameters['node']->hasField('field_member_of') &&
```
whikloj: If you are creating a configuration that (I assume) will be editable, maybe the field we traverse should also be configurable. For now if you add it to the same config/install file we should be good.

modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
whikloj: I'm just noting that you mentioned you were not including a default config. I'd do that as it will be required once you try to create a functional test.

seth-shaw-unlv: now including a default config

modules/islandora_breadcrumbs/src/IslandoraBreadcrumbBuilder.php
seth-shaw-unlv: The reference field is now configurable.

seth-shaw-unlv:
@whikloj we could do that. First we need to setup a bundle that includes the reference field (since repository object is defined in islandora_defaults, not islandora itself), then we can setup a chain to ensure the normal mode works. We can then loop the chain to make sure we escape it properly.

seth-shaw-unlv:
Ah, I see IslandoraFunctionalTestBase already does setup the content type and field. Good, good.

whikloj:
Testing

whikloj:
We do need to enable the tests for this, and we need a form to adjust the configurable settings in here.

whikloj:
@Islandora-CLAW/committers do we want the internal tests enabled here before I merge. I started work to make enabling the tests easier (https://github.com/Islandora-CLAW/CLAW/pull/1146, https://github.com/Islandora-CLAW/islandora/pull/142).

We can just merge this and worry about enabling the tests later, thoughts?

.travis.yml
```diff
@@ -27,7 +27,7 @@ install:
 
 script:
   - $SCRIPT_DIR/travis_scripts.sh
-  - $SCRIPT_DIR/run-tests.sh "islandora"
+  - $SCRIPT_DIR/run-tests.sh "islandora_breadcrumbs"
```
whikloj: Sorry I meant add a new line, this will disable the islandora tests which we still need.

whikloj: I am not being clear, this line is fine. but don't remove the other `run-tests.sh "islandora"`

seth-shaw-unlv: D'oh; I should have caught that anyway.

PR #133, opened 2019-05-14
Respecting jsonld module configuration for ?_format=jsonld

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/887

Goes with
- https://github.com/Islandora-CLAW/jsonld/pull/32
- https://github.com/Islandora-CLAW/Crayfish/pull/65

# What does this Pull Request do?

Forces our jsonld alters to respect the `jsonld` module's configuration for stripping out `?_format=jsonld` from urls.

# What's new?

The jsonld module's configuration is injected into our base class for altering jsonld using context.  There's a helper function implemented in the base class that generates the appropriate url to use when searching the jsonld graph for the entity in question.

# How should this be tested?

See https://github.com/Islandora-CLAW/CLAW/issues/887

# Interested parties
@Islandora-CLAW/committers

PR #134, opened 2019-05-15
 Respecting jsonld module configuration for ?_format=jsonld

dannylamb:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/887

Goes with
- https://github.com/Islandora-CLAW/jsonld/pull/33
- https://github.com/Islandora-CLAW/Crayfish/pull/66

Supersedes: https://github.com/Islandora-CLAW/islandora/pull/133

**This is the same code as https://github.com/Islandora-CLAW/islandora/pull/133, just on the Foundation's repo so that it can be tested with claw-playbook**

Forces our jsonld alters to respect the `jsonld` module's configuration for stripping out `?_format=jsonld` from urls.

# What's new?

The jsonld module's configuration is injected into our base class for altering jsonld using context.  There's a helper function implemented in the base class that generates the appropriate url to use when searching the jsonld graph for the entity in question.

# How should this be tested?

See https://github.com/Islandora-CLAW/CLAW/issues/887

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
   "require": {
     "drupal/context": "^4.0",
     "drupal/search_api": "~1.8",
-    "islandora/jsonld": "dev-issue-887",
+    "islandora/jsonld": "0.2.0",
     "stomp-php/stomp-php": "4.*",
     "drupal/jwt": "1.0.0-alpha6",
     "drupal/filehash": "^1.1",
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
   "require": {
     "drupal/context": "^4.0",
     "drupal/search_api": "~1.8",
-    "islandora/jsonld": "dev-8.x-1.x",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
   "require": {
     "drupal/context": "^4.0",
     "drupal/search_api": "~1.8",
-    "islandora/jsonld": "dev-8.x-1.x",
+    "islandora/jsonld": "dev-issue-887",
```
whikloj: Once an @Islandora-CLAW/committers merges https://github.com/Islandora-CLAW/jsonld/pull/33 (@dannylamb and I have commits in it). Then this needs to be changed.

whikloj: We should cut a `0.2.0` release of jsonld and then change this to that.

PR #137, opened 2019-05-23
Forcing flysystem downloads to be language neutral

dannylamb:
**GitHub Issue**: Resolves https://github.com/Islandora-CLAW/CLAW/issues/985

# What does this Pull Request do?

Forces flysystem downloads from Fedora to be language neutral.  This prevents the language prefix from getting jammed into the url.

# What's new?
An overridden getExternalUrl() method that forces the language to be Undefined (und).

# How should this be tested?

- Make a second language and set one to default 
- Go visit a media that you put in Fedora, and its file's url should have the language prefix in it, e.g. `http://localhost:8000/en/_flysystem/fedora/2019-05/blah.jpg`
- Pull in this PR
- Clear the cache
- Go visit the same media, and now the language prefix should be gone, e.g. `http://localhost:8000/_flysystem/fedora/2019-05/blah.jpg`

# Interested parties
@kanasznagyzoltan @rangel35 @DonRichards @Islandora-CLAW/committers

PR #141, opened 2019-05-29
Fixes content translation

dannylamb:
**GitHub Issue**: Resolves https://github.com/Islandora-CLAW/CLAW/issues/1127

The same fix for https://github.com/Islandora-CLAW/islandora/pull/137, but applied to urls getting put into messages on their way to the queue.  This may also partially resolve issues arising from https://github.com/Islandora-CLAW/CLAW/issues/1095.  At least this same pattern will have to be repeated (read: put into utility function and just call that).

# What's new?
Forcing URLs to be language neutral.  Plus some minor refactoring to properly inject the language manager.

# How should this be tested?

Confirm Original Behaviour:
- Add a language to turn on content translation (I added French)
- Make a node
- Click its translate tab
- Add a translation (again, French for me)
- Save the translation
- Go check out the node in the triplestore and fedora, there's no French metadata.

Confirm fixed behaviour
- Pull in this PR
- `drush cr`
- Go edit the translation and save it
- Go check out the node in the triplestore and fedora, and you should see both French and English metadata.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/islandora.services.yml
+++ b/islandora.services.yml
@@ -3,6 +3,7 @@
 services:
   islandora.eventgenerator:
     class: Drupal\islandora\EventGenerator\EventGenerator
+    arguments: ['@language_manager']
   islandora.stomp:
     class: Stomp\StatefulStomp
     factory: ['Drupal\islandora\StompFactory', create]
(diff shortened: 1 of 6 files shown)


Natkeeran:
This looks good.

PR #146, opened 2019-06-23
Adds a Context Condition that tests a node's parent

mjordan:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1170

Adds a Context Condition that tests a node's parent, which can be either a collection object or not.

# What's new?

Adds the new Condition, and also adds its ID to the list in `islandora.module` that excludes this condition from the Block placement UI.

# How should this be tested?

1. Check out this branch and `drush cr`.
1. Create a new Context using the "Node has parent" condition.
1. In the Parent Node field, enter the name of a collection.
1. Choose the name of the field that contains the parent reference (probably `field_member_of`)
1. Use "Theme" as the Reaction, and choose the Bartik theme.
1. Go to an object that is not in the collection you named in your Conext. The theme should remain the default.
1. Now go to an object that is in the collection you named in your Conext. The theme should change to Bartik.
1. Go back to the object not in the collection.
1. To test membership of a complex object, add an object to another non-collection object as a child.
1. Reconfigure your context's condition to indicate the non-collection object as the parent.
1. Visit the child object. The theme should change to Bartik.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/config/schema/islandora.schema.yml
+++ b/config/schema/islandora.schema.yml
@@ -87,10 +87,10 @@ condition.plugin.node_has_parent:
   type: condition.plugin
   mapping:
     parent_nid:
-      type: ignore
+      type: integer
       label: 'Parent node'
     parent_reference_field:
-      type: ignore
+      type: string
       label: 'Parent reference field'
 
 condition.plugin.media_has_term:
--- a/config/schema/islandora.schema.yml
+++ b/config/schema/islandora.schema.yml
@@ -83,6 +83,16 @@ condition.plugin.node_has_term:
       type: text
       label: 'Taxonomy Term URI'
 
+condition.plugin.node_has_parent:
+  type: condition.plugin
+  mapping:
+    parent_nid:
+      type: ignore
+      label: 'Parent node'
+    parent_reference_field:
+      type: ignore
+      label: 'Parent reference field'
+
 condition.plugin.media_has_term:
   type: condition.plugin
   mapping:
--- a/src/Plugin/Condition/NodeHasParent.php
+++ b/src/Plugin/Condition/NodeHasParent.php
@@ -96,7 +96,7 @@ class NodeHasParent extends ConditionPluginBase implements ContainerFactoryPlugi
       '#options' => $options,
       '#default_value' => $this->configuration['parent_reference_field'],
       '#required' => TRUE,
-      '#description' => t("Machine field name that contains references to parent node. You normally do not need to change this value."),
+      '#description' => t("Machine field name that contains references to parent node."),
     ];
 
     return parent::buildConfigurationForm($form, $form_state);
--- a/src/Plugin/Condition/NodeHasParent.php
+++ b/src/Plugin/Condition/NodeHasParent.php
@@ -89,7 +89,8 @@ class NodeHasParent extends ConditionPluginBase implements ContainerFactoryPlugi
       '#target_type' => 'node',
     ];
     $field_map = \Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference');
-    $options = array_keys($field_map['node']);
+    $node_fields = array_keys($field_map['node']);
+    $options = array_combine($node_fields, $node_fields);
     $form['parent_reference_field'] = [
       '#type' => 'select',
       '#title' => t('Field that contains reference to parents'),
--- a/src/Plugin/Condition/NodeHasParent.php
+++ b/src/Plugin/Condition/NodeHasParent.php
@@ -88,9 +88,12 @@ class NodeHasParent extends ConditionPluginBase implements ContainerFactoryPlugi
       '#description' => t("Can be a collection node or a compound object."),
       '#target_type' => 'node',
     ];
+    $field_map = \Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference');
+    $options = array_keys($field_map['node']);
     $form['parent_reference_field'] = [
-      '#type' => 'textfield',
+      '#type' => 'select',
       '#title' => t('Field name that contains reference to parents'),
+      '#options' => $options,
       '#default_value' => $this->configuration['parent_reference_field'],
       '#required' => TRUE,
       '#description' => t("Machine field name that contains references to parent node. You normally do not need to change this value."),
--- a/src/Plugin/Condition/NodeHasParent.php
+++ b/src/Plugin/Condition/NodeHasParent.php
@@ -1,0 +1,164 @@
+<?php
+
+namespace Drupal\islandora\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a condition to detect node's parent.
+ *
+ * @Condition(
+ *   id = "node_has_parent",
+ *   label = @Translation("Node has parent"),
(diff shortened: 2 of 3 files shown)


src/Plugin/Condition/NodeHasParent.php
```diff
@@ -0,0 +1,164 @@
...
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['parent_nid'] = [
+      '#type' => 'entity_autocomplete',
+      '#title' => t('Parent node'),
+      '#default_value' => $this->entityTypeManager->getStorage('node')->load($this->configuration['parent_nid']),
+      '#required' => TRUE,
+      '#description' => t("Can be a collection node or a compound object."),
+      '#target_type' => 'node',
+    ];
+    $form['parent_reference_field'] = [
```
seth-shaw-unlv: I don't suppose we could make this a select box and build our list from the available entity reference fields that are available to the site? Something like:
```
    $form['parent_reference_field'] = [
      '#type' => 'select',
      '#title' => t('Field name that contains reference to parents'),
      '#options' => array_keys(\Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference')),
      '#default_value' => $this->configuration['parent_reference_field'],
      '#required' => TRUE,
      '#description' => t("Machine field name that contains references to parent node. You normally do not need to change this value."),
    ];
```

mjordan: Possibly... I'll give it a try.

mjordan: OK, now we have:

reference_fields

seth-shaw-unlv:
This new Context is throwing an error when I try to view an item to which the item applies: `[Tue Jun 25 16:20:17.690539 2019] [php7:notice] [pid 29293] [client 10.0.2.2:53599] Uncaught PHP Exception InvalidArgumentException: "Field 5 is unknown." at /var/www/html/drupal/web/core/lib/Drupal/Core/Entity/ContentEntityBase.php line 586, referer: http://localhost:8000/node/4/edit`

Ah, I think the select broke it somehow. My new Context's config is:
```
uuid: ac6877ce-fb6c-4561-9554-20106d15cc8c
langcode: en
status: true
dependencies:
  module:
    - islandora
name: node_belongs_to_collection
label: 'Node belongs to Collection'
group: Islandora
description: ''
requireAllConditions: false
disabled: false
conditions:
  node_has_parent:
    id: node_has_parent
    negate: 0
    parent_reference_field: '5'
    parent_nid: '1'
    uuid: ed09666b-5891-44ba-8a5d-c800184121f4
    context_mapping:
      node: '@node.node_route_context:node'
reactions:
  theme:
    id: theme
    theme: bartik
    saved: false
    uuid: eb1ffd3e-ad2b-43cf-9b79-856281556a00
weight: 0
```
Notice how parent_reference_field is '5' instead of 'field_member_of'. Looks like that options array needs to be fixed somehow.

seth-shaw-unlv:
According to [the docs for selects](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Select.php/class/Select/8.7.x), the '#options' value should be:
> An associative array, where the keys are the values for each option, and the values are the option labels to be shown in the drop-down list.

Since the options array is not associative, and field_member_of was the fifth in the list, the context stored '5' as the field name.

src/Plugin/Condition/NodeHasParent.php
seth-shaw-unlv: Looks like I mislead you here. We need to alter that options array into an associative one.

mjordan:
Thanks @seth-shaw-unlv, I'll take a look at this this evening.

src/Plugin/Condition/NodeHasParent.php
seth-shaw-unlv:
```
$field_map = \Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference');
$node_fields = array_keys($field_map['node']);
$options = array_combine($node_fields, $node_fields);
```

seth-shaw-unlv:
I mentioned this in the resolved code thread, but it isn't appearing on this page, this will fix it:
```
diff --git a/src/Plugin/Condition/NodeHasParent.php b/src/Plugin/Condition/NodeHasParent.php
index 96ae8f2..df95c6a 100644
--- a/src/Plugin/Condition/NodeHasParent.php
+++ b/src/Plugin/Condition/NodeHasParent.php
@@ -89,7 +89,8 @@ class NodeHasParent extends ConditionPluginBase implements ContainerFactoryPlugi
       '#target_type' => 'node',
     ];
     $field_map = \Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference');
-    $options = array_keys($field_map['node']);
+    $node_fields = array_keys($field_map['node']);
+    $options = array_combine($node_fields, $node_fields);
     $form['parent_reference_field'] = [
       '#type' => 'select',
       '#title' => t('Field that contains reference to parents'),
```

src/Plugin/Condition/NodeHasParent.php
```diff
@@ -0,0 +1,167 @@
...
+      '#type' => 'entity_autocomplete',
+      '#title' => t('Parent node'),
+      '#default_value' => $this->entityTypeManager->getStorage('node')->load($this->configuration['parent_nid']),
+      '#required' => TRUE,
+      '#description' => t("Can be a collection node or a compound object."),
+      '#target_type' => 'node',
+    ];
+    $field_map = \Drupal::service('entity_field.manager')->getFieldMapByFieldType('entity_reference');
+    $options = array_keys($field_map['node']);
+    $form['parent_reference_field'] = [
```
dannylamb: Gonna need a schema entry for this one, too.  Not sure how you tell Drupal it's an entity reference, though.  I'm thinking you say it's an Integer because it stores node IDs, but that's just a guess.

dannylamb: Ah, seeing now that this particular line is a field reference...  another kink.  I guess the node reference is an int and the field reference is a string if we're talking about storing ids.  [You have a pretty limited choice of things to work with](https://www.drupal.org/docs/8/api/configuration-api/configuration-schemametadata#types).

mjordan: @dannylamb I'll use `ignore` since that is what seems most applicable.

config/schema/islandora.schema.yml
```diff
@@ -83,6 +83,16 @@ condition.plugin.node_has_term:
       type: text
       label: 'Taxonomy Term URI'
 
+condition.plugin.node_has_parent:
+  type: condition.plugin
+  mapping:
+    parent_nid:
+      type: ignore
```
seth-shaw-unlv: should be integer.

config/schema/islandora.schema.yml
```diff
@@ -83,6 +83,16 @@ condition.plugin.node_has_term:
...
+      label: 'Parent node'
+    parent_reference_field:
+      type: ignore
```
seth-shaw-unlv: should be string.

src/Plugin/Condition/NodeHasParent.php
seth-shaw-unlv: @mjordan, @dannylamb is right. the parent_pid and parent_reference field are 'integer' and 'string' respectively.

config/schema/islandora.schema.yml
mjordan: OK, will update.

config/schema/islandora.schema.yml
mjordan: OK, will update.

mjordan: Done.

config/schema/islandora.schema.yml
mjordan: Done.

PR #154, opened 2019-07-25
Versioning

elizoller:
**GitHub Issue**: Related to: (https://github.com/Islandora-CLAW/CLAW/issues/740)

* Allows versioning of an object to be initiated from Drupal node updates

# What's new?
* Add a isNewVersion boolean value to the event object that gets sent to Alpaca

# How should this be tested?
* pull in the PR to your islandora modules
* make sure that the event type is included in the emitted event object (could debug from EventGenerator or update an alpaca feature to try this)

# Additional Notes:
One important note is that using the IndexReaction for triggering the create_node_version_in_fedora means that the event is triggered as part of islandora_node_insert and islandora_node_update (i think...). Having it be part of node_update means that the version is created before the node is updated. 
Another way to do this is to use a new reaction (maybe called VersionReaction) that could then be included in any hooks in `islandora.module` although probably islandora_node_update makes the most sense.

# Interested parties
@Islandora-CLAW/committers


src/EventGenerator/EventGenerator.php
```diff
@@ -99,6 +99,8 @@ public function generateEvent(EntityInterface $entity, UserInterface $user, arra
       $event["summary"] = ucfirst($data["event"]) . " a " . ucfirst($entity_type);
     }
 
+    $event["object"]["type"] = $event["type"];
```
dannylamb: You should check for versioning in the if statement above and just use that logic.  E.g. your event type will be "activity", and summary will be "version".

dannylamb: Or maybe not... see my comment below.

dannylamb:
@elizoller Doing some digging here, and it looks like the only way you can create a revision in Drupal is by saving a node, even programatically...  So a revision is always tied to the current state of the entity, even if nothing's changed except the version id.  If we fire off two events, one for save, and one for revisioning, there's no guarantee those things are going to line up or be processed in the correct order.  Maybe you can just jam a boolean into the indexing event to signal we want to create a revision?

elizoller:
The way i was understanding it is that the new "create_node_version_in_fedora" action uses the plugin "emit_node_event" which is the EmitNodeEvent class (in src/Plugin/Action/EmitNodeEvent.php) which extends EmitEvent class (in src/EventGenerator/EmitEvent.php) which when executed calls `$this->eventGenerator->generateEvent(...)` which now passes along the `$event['type']` (which is which configuration:event in the "create_node_version_in_fedora" action. and the "create_node_version_in_fedora" action must be selected in the context for repository item nodes (admin/structure/context/repository_content) under the "Index" reaction. having that pre-selected could ship as part of islandora_defaults if we think most people would want versioning in fedora to happen automatically.

elizoller:
so it sounds like we've concluded it isn't necessary to have a separate reaction for versioning that is also associated with hook_node_update?

dannylamb:
@elizoller,  I'm still playing with it, but yes. I think your best bet is to just make the the index/create events contain whether or not to make a version.  Since you have to save the entity to make the version in Drupal, they're not really all that independent of events.

whikloj:
@elizoller I did not get to this exchange before I sent you my comments [here](https://github.com/Islandora-CLAW/Crayfish/pull/74#discussion_r311158943). Maybe you should just fix the header name if @dannylamb thinks a new endpoint is not worth the trouble.

elizoller:
Haha, already did the work... but I think you are kind of talking about different things. @whikloj, i think you were suggesting having different endpoints in milliner and not using a header in the Alpaca/Milliner section where as i think @dannylamb is more focused on not getting the events out of sync on the Drupal-side, so keeping the Version and saveNode events emitting from the same hook and stream.

elizoller:
Based on discussion at the Tech call today, we want versioning to be fired in Milliner in an idempotent way - ie. it can fired again and again but if its the same "version" from Drupal, we don't want to create multiple duplicate versions in Fedora. Thus, there does not need to be a separate event emitted (emit_node_event) into the node queue (islandora-indexing-fcrepo-content). So in Drupal, it would be more of a matter of allowing a user to configure on what content versions are created in Fedora - ie. on what content type or media type. What would be the best place to put this? I was thinking the Islandora Settings page (much like the Gemini URI Psuedo field) but I'm not sure if there is a more Drupal-y place to put it. Thoughts?
Also that would probably mean this PR would be closed without being merged and I'd put up a new PR with some other config piece.

elizoller:
@whikloj @dannylamb After discussion at the tech call yesterday, there was concern about having the call to Fedora to create a version be idempotent so we don't end up with something like this (where each version is the same "revision" in Drupal and thus the same RDF):
Screen Shot 2019-08-08 at 11 12 33 AM

- Islandora: We need some way to pass the fact that we want to make a version from Drupal all the way Milliner (via Alpaca). The method here was to create a separate Action that got fired when a node was saved but if we go the route below with chaining the calls to Milliner, then I don't think we need a separate action. What we need is some way to tell if we want to create a version and pass that from the EventGenerator through to Alpaca. I wonder if it should be a configuration by bundle type in the Islandora settings (like the Gemini URI psuedo-field) - this might be the easiest option? Or if the most logical place to keep it is in context, then maybe the action should just not use the emit_node_event plugin (although I'm not entirely clear on if/how we would tell from one action whether or not another action is being called from the same node...) Or I'm open to other options on a Drupal-y way to do this.

- Alpaca: If we want to keep Milliner with its own versioning endpoint (see justification below), then I would probably just change the [when conditional](https://github.com/Islandora-CLAW/Alpaca/pull/61/files#diff-a0a4e90c5cf5024700844b61bbe9e12eR137) so that it doesn't have an "otherwise" and always calls the regular save/update milliner endpoint but only conditionally calls the versioning endpoint if Drupal tells it to. Unless you all think chaining the two commands either doesn't actually work or is not good practice.

- Milliner: In my opinion it's cleaner to separate the versioning functionality in Milliner from the saving of node. I think this is beneficial because the 'version/uuid' endpoint could be reused across nodes and media objects (I think). Also, I think it lends itself to more extensibility down the road. I will add to the current Milliner PR the piece that looks at the date modified coming from Drupal and tries not to make repeated calls (such as how the MillinerService->updateNode works) so we can make the version endpoint idempotent.

Obviously I'm totally open to discuss all of this and am super grateful for all of the feedback you all have provided!

dannylamb:
@elizoller I'd just put an if statement using [isNewRevision](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21RevisionableInterface.php/function/RevisionableInterface%3A%3AisNewRevision/8.7.x) in the [EventGenerator]() class somewhere around [here](https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/src/EventGenerator/EventGenerator.php#L101) to jam the results of that function into the event.  Then we'll respect versions no matter what the user does, using context or some other means to trigger versions.

I think the rest of what you're suggesting is sound.  The version endpoint in Milliner can just call the same update/save code and then slap the function to make a new version on the end.  To make sure the version function itself can get called over and over again, you may very well be able to lift that functionality from other parts of Milliner.  Look [here](https://github.com/Islandora-CLAW/Crayfish/blob/dev/Milliner/src/Service/MillinerService.php#L214-L260) where it's checking modified dates.  There's a lot tangled up in that chunk, but you should be able to separate it out or re-use the code somehow.

elizoller:
By "Then we'll respect versions no matter what the user does, using context or some other means to trigger versions." - are you saying that we should tell Fedora to always make a version and not let the user configure in Drupal whether or not that happens?

dannylamb:
I'm saying we should make it respect what the user does, however they want to do it.  If they make a new version in Drupal, they'll get a new version in Fedora.  Otherwise, it'll just save it to Fedora with no new version getting made.

It's up to them to determine when/how to slice new versions in Drupal, either through auto-versioning or using something like context.

elizoller:
Oh! Now it clicked. isNewRevision will only be fired if they enable Drupal from the content type settings "Create new revision"

elizoller:
isNewVersion is deprecated and replaced with shouldCreateNewRevision
(https://api.drupal.org/api/drupal/core%21modules%21node%21src%21Entity%21NodeType.php/function/NodeType%3A%3AisNewRevision/8.7.x)

elizoller:
big thank you to @qadan and @whikloj for helping me through this

seth-shaw-unlv:
@elizoller, can you update the testing instructions? There is no longer an action to add to the context. Thanks.

elizoller:
For testing, you only need to pull this PR into your regular islandora module. The versioning happens automatically in the EventGenerator for bundles which are versionable and have shouldCreateNewRevision = true

seth-shaw-unlv:
This appears to work for me. 👍

The one slightly odd bit to me is that we are creating a version when the node is first inserted, but then we make another version of the same data before we do an update. So the first two versions of a given resource will always be the same. It seems to me that, if we are versioning content before updating the resource, we _shouldn't_ a version on node insert; just node update.

elizoller:
That makes sense to me conceptually, I thought perhaps I could look at the `$event['type']` or `$event['summary']` in the EventGenerator but that doesn't seem to be telling me if the node is being created or updated. Do you know where I might find that kind of information?

seth-shaw-unlv:
🤷‍♂️, not sure. /me digs through docs....  What would happen if we used [isNewRevision](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21RevisionableInterface.php/function/RevisionableInterface%3A%3AisNewRevision/8.8.x) on the entity rather than testing the bundle type?

elizoller:
shouldCreateNewRevision isn't available on an entity directly as far as I can tell 
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21RevisionableEntityBundleInterface.php/function/RevisionableEntityBundleInterface%3A%3AshouldCreateNewRevision/8.4.x ?

seth-shaw-unlv:
That is why I mentioned 'isNewRevision' which is available to a revisionable entity.

elizoller:
I thought isNewRevision is deprecated? https://api.drupal.org/api/drupal/core%21modules%21node%21src%21Entity%21NodeType.php/function/NodeType%3A%3AisNewRevision/8.7.x

seth-shaw-unlv:
[isNewRevision for _NodeType_](https://api.drupal.org/api/drupal/core%21modules%21node%21src%21Entity%21NodeType.php/function/NodeType%3A%3AisNewRevision/8.8.x) is deprecated; but [isNewRevision for _RevisionableInterface_](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21RevisionableInterface.php/function/RevisionableInterface%3A%3AisNewRevision/8.8.x) (which I believe Node implements) isn't deprecated.

elizoller:
gotcha, thanks for the clarification on the deprecation.
so in theory, i could call isNewRevision right off the entity object then?
I tried ```$isNewRev = FALSE;
    if ($entity->getEntityType()->isRevisionable()) {
      \Drupal::logger('islandora')->notice('its revisionable');
      $isNewRev = $entity->isNewRevision();
    }
    $event["object"]["isNewVersion"] = $isNewRev;```
and i'm able to see the log that the entity is revisionable but i'm not getting isNewRevision to be true on node creation or node update. I'm thinking I'm missing something obvious here about Drupal 8 programming - thoughts?

seth-shaw-unlv:
Did a little digging. We fire these events based on [hook_node_insert](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_ENTITY_TYPE_insert/8.8.x) and [hook_node_update](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_ENTITY_TYPE_update/8.8.x) which only fire _after_ "the entity storage has been updated". Well, ContentEntityBase uses the isNewRevision as part of it's serialization process and sets this value to FALSE during [postSave](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityBase.php/function/ContentEntityBase%3A%3ApostSave/8.8.x) which is executed _before_ our hooks. So, I sent us barking up the wrong tree.

One thing we could do, is check and see if a node has more than one revision `$hasMultipleRevisions = (count(\Drupal::entityManager()->getStorage($entity-> getEntityType())->revisionIds($entity)) > 1)`. If so, we know this is an update (assuming revisions for this entity have been enabled) because creating a node makes the first revision.

elizoller:
Technically, I got this to work with some slight modifications to the snippet you provided above but conceptually I'm feeling uneasy about it. An early version of this had be creating a separate reaction that got fired for versioning and it seems to me like it would be more logical to circle back to that kind of approach. Ie. a separate reaction that is fired on postSave that would kick off a versioning call (this would likely affect how the call is separated out in Alpaca).

elizoller:
Also, i'm waiting to push up because I'm seeing if I'm supposed to push to the migrated PR at https://github.com/Islandora/islandora/pull/733 but I currently don't have permissions

elizoller:
@seth-shaw-unlv lets move discussion here: https://github.com/Islandora/islandora/pull/733

elizoller:
can i close this PR since https://github.com/Islandora/islandora/pull/733 has been merged?

dannylamb:
@elizoller Please do.

elizoller:
closing this since https://github.com/Islandora/islandora/pull/733 has been merged

PR #157, opened 2019-09-04
Initial implementation of a REST Export serializer for IIIF Manifests

joecorall:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1263

Create a plugin for generating an IIIF manifest from views results.

# What's new?
Adds a submodule `islandora_iiif` that has a Views style plugin for a serializer that prints IIIF URLs for images displayed in a View.

# How should this be tested?

* Enable `islandora_iiif` submodule
* Add your IIIF Server URL (the same one you'd have configured for OSD) at `/admin/config/islandora/iiif`
* Import this "test" // "demo" view: https://d.pr/f/DV5GID
* Access a node that has an image referencing the node via `field_media_of` e.g. /node/1
* Add "iiif-manifest" to the URL e.g. /node/1/iiif-manifest
* You should see the JSON of every image for the respective node

# Additional Notes:
For the Paged Content Sprint

# Interested parties
@Islandora-CLAW/committers


modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -0,0 +1,198 @@
...
+
+      /** @var \Drupal\Core\Field\FieldItemListInterface $images */
+      $images = $entity->{$viewsField->definition['field_name']};
+      foreach ($images as $image) {
+        $file = $image->entity;
+        $resource = $this->fileinfoService->getFileData($file);
+
+        // remove $base_url from full_path
+        $path = $resource['full_path'];
+        $path = str_replace($base_url, '', $path);
```
whikloj: I don't think we want to do this, we started to pass the full URL to the image in OSD [here](https://github.com/Islandora-CLAW/openseadragon/commit/db23c1c66c0bf0234724e159adb76a58be8c0813) it also requires a change to the IIIF server config (here in [claw-playbook](https://github.com/Islandora-Devops/claw-playbook/commit/f23aa2f294d78f8417bc8f0b233803eb2b1992e3) or [here](https://github.com/Islandora-Devops/ansible-role-cantaloupe/blob/master/defaults/main.yml#L167) in the actual role), but then (in theory) you can have Drupal and Cantaloupe on separate servers

whikloj:
Travis is failing on style errors, if you are running the claw-playbook you should be able to duplicate them with
```
/var/www/html/drupal/vendor/bin/phpcs --standard=/var/www/html/drupal/vendor/drupal/coder/coder_sniffer/Drupal /var/www/html/drupal/web/modules/contrib/islandora/modules/islandora_iiif
```
and then fix most of them with
```
/var/www/html/drupal/vendor/bin/phpcbf --standard=/var/www/html/drupal/vendor/drupal/coder/coder_sniffer/Drupal /var/www/html/drupal/web/modules/contrib/islandora/modules/islandora_iiif
```
I think my directories are correct.

mjordan:
Details on phpcs are available at https://github.com/Islandora-CLAW/CLAW/blob/master/docs/technical-documentation/checking-coding-standards.md.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
kayakr: Just wanted to mention that we have Cantaloupe on a different server than Islandora, for performance (memory) reasons.

joecorall: Just so I'm clear, so instead of having the IIIF URL's formatted like this:

`http://localhost::8080/cantaloupe/iiif/2/sites%2Fdefault%2Ffiles%2Fbepress%2Fwomensleadership_2017_bio_gallery%2F1007%2FDeniseheadshot2.jpg`

the IIIF URL's should output like this:

`http://localhost::8080/cantaloupe/iiif/2/http%3A%2F%2Flocalhost%3A8000%2Fsites%2Fdefault%2Ffiles%2Fbepress%2Fwomensleadership_2017_bio_gallery%2F1007%2FDeniseheadshot2.jpg`

I had customized my Cantaloupe config to use the FileSystem resolver instead of HTTP, so will just need to revert my customizations back to the claw-playbook.

joecorall: Answered my own question by updating OSD to latest version. Commit coming soon that will resolve this. Thanks for catching it!

seth-shaw-unlv:
Also, correct me if I'm wrong @dannylamb, but weren't we going to slot this directly into the module rather than creating sub-modules? If so, then `modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php` can be moved to `src/Plugin/views/style/IIIFManifest.php` (with the namespace updated accordingly) and the other two files deleted.

seth-shaw-unlv:
Oh, [never mind](https://islandora.slack.com/archives/CM5PPAV28/p1567622919286800?thread_ts=1567536699.254000&cid=CM5PPAV28).

dannylamb:
re: the dependencies, it seems like the only things getting used from openseadragon are the iiif url and the `FileInformation` utility class.  i'm racking my brain trying to see if we can move those around wihtout introducing a dependency on islandora in openseadragon, which we're purposefully trying to keep separate.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -0,0 +1,261 @@
...
+      $container->get('openseadragon.fileinfo')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
+    $json = [];
+    $viewer_settings = $this->openseadragonConfig->getSettings(TRUE);
```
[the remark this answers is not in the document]
joecorall: Nice catch. I removed it.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -0,0 +1,261 @@
...
+    if (isset($entity->{$viewsField->definition['field_name']})) {
+
+      /** @var \Drupal\Core\Field\FieldItemListInterface $images */
+      $images = $entity->{$viewsField->definition['field_name']};
+      foreach ($images as $image) {
+        // Create the IIIF URL for this file
+        // Visiting $iiif_url will resolve to the info.json for the image.
+        $file = $image->entity;
+        $resource = $this->fileinfoService->getFileData($file);
+        $iiif_url = rtrim($iiif_address, '/') . '/' . urlencode($resource['full_path']);
```
dannylamb: You can call `$file->url()` here instead of using the fileInfoService.  That would eliminate a dependency.

joecorall:
> re: the dependencies, it seems like the only things getting used from openseadragon are the iiif url and the `FileInformation` utility class. i'm racking my brain trying to see if we can move those around wihtout introducing a dependency on islandora in openseadragon, which we're purposefully trying to keep separate.

Yeah, I think one of those modules will need to "own" the IIIF URL. That is what I saw as the major dependency between islandora <-> openseadragon. The only way to avoid an explicit dependency inside of info.yml I think is for one of those modules to use something like `\Drupal::service('module_handler')->moduleExists('openseadragon')` and if the other module exists, read the other module's config variable, and if the module doesn't exist to define the IIIF URL config variable itself.

dannylamb:
@joecorall Agreed, the IIIF url is the real dependency wrecker.  I think it'd be ok to have the iiif submodule own the URL and have openseadragon do a soft check for it.  A bit out of scope for this, i think.  We can always tackle it when we adapt the openseadragon viewer to use these manifests.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
joecorall: Fixed with latest commit (which also puts the IIIF Server URL into the islandora_iiif module, removing the OSD dependency on islandora_iiif)

dannylamb:
@joecorall I'm spinning up an environment now, and will give it a final test when done.  Let's get this in and move on to the viewer :rocket:

dannylamb:
@joecorall This is going to need a schema entry for the settings config.  Something like this should do (blatant copy/pasta from the core `islandora_module`):

```
islandora_iiif.settings:
  type: config_object
  label: 'Islandora IIIF Settings'
  mapping:
    islandora_iiif:
      type: string
      label: 'IIIF Server Url'
```

dannylamb:
@joecorall Attempting to enable the module is giving me this:

`islandora_iiif.settings key contains a dot which is not supported.`

dannylamb:
@joecorall I see what's going on now.  What's currently in `config/install/islandora_iiif.settings.yml` needs to be in `config/schema/islandora_iiif.schema.yml` and you can revert that file back to just `iiif_server: `

joecorall:
Thanks. I always generate the code for config forms/settings using `drupal generate:form:config`, and it seems to leave off those schema.yml settings. Should have read the docs closer.

dannylamb:
@joecorall It's harmless up until you try and write a test and then blammo, schema config errors everywhere.  Just one of those things I've learned to catch up front.

dannylamb:
@joecorall Ok, finally putting this through its paces.  I'm trying to set up a view that gets all the original files for members of a node.  Everything's going good, except when I try to set the source field on the IIIF Manifest format, I'm getting an empty dropdown.

Here's what my view looks like:
Screenshot from 2019-09-11 11-45-31

I think if I can just tell that to look for `field_media_file` or `field_media_image`, we should be good to go on this.

And here's the empty drop down:

Screenshot from 2019-09-11 11-44-20

joecorall:
I have an example view here >>> [https://d.pr/f/DV5GID+](https://d.pr/f/DV5GID+) <<< right now I believe that dropdown would only show "image" fields as options. Though I guess a TIFF might be represented as a "file", not an "image". Let me take a look at how to handle "Media: File"

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -0,0 +1,249 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
+    parent::buildOptionsForm($form, $form_state);
+
+    $field_options = [];
+
+    $fields = $this->displayHandler->getHandlers('field');
```
dannylamb: This doesn't seem to be getting the file and image fields off of our media types.  Will have to keep digging, but I would expect `field_media_file` and `field_media_image` to be in the dropdown here.

dannylamb:
The view is for Media and not Nodes, maybe that's a monkey wrench?  In order to get all Original Files for all nodes who belong to a collection/book, you have to make a media view and then set the media_of relationship.  Then you can use the node id of collection/book as a contextual filter.  There's a few layers of indirection going on there.

joecorall:
Please pull down the latest commit. You should now have the option of setting "file" fields as the source field. I don't like the approach I took to see which fields in a View are images or files, but this new approach does ensure `field_media_file` and `file_media_image` are always available as options, and it attempts to make sure custom file/image fields are available, too. I can't find a straightforward way to see if the View field is referencing an image or file, so it looks for the [View Field formatter](https://d.pr/i/AmwAGu) for a known formatter for files/images.

dannylamb:
I just clobbered my environment to test out https://github.com/Islandora-CLAW/CLAW/issues/1269.  I'll test it as soon as it spins up again.

dannylamb:
:+1: :rocket:

Yep, I can now make media views and select tiffs using the file field.  Let's do this @joecorall!

:shipit:

joecorall:
Glad to help. Thanks for everyone on this thread for their testing/reviewing/patience!!

PR #158, opened 2019-09-05
Fixes PHP warning by testing pseudo_bundles before iterating.

kayakr:
**GitHub Issue**:
https://github.com/Islandora-CLAW/CLAW/issues/1266

Fix PHP warning arising from no selection of bundles with Gemini URI pseudo field.
http://localhost:8000/admin/config/islandora/core

# How should this be tested?

We're seeing this warning on pre-1.0.0 instances. It doesn't appear to happen on a fresh VM since at least some bundles are selected. With no bundles selected warning does not occur, even with no patch.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/islandora.module
+++ b/islandora.module
@@ -362,14 +362,16 @@ function islandora_entity_extra_field_info() {
 
   $pseudo_bundles = $config_factory->get(IslandoraSettingsForm::GEMINI_PSEUDO);
 
-  foreach ($pseudo_bundles as $key) {
-    list($bundle, $content_entity) = explode(":", $key);
-    $extra_field[$content_entity][$bundle]['display']['field_gemini_uri'] = [
-      'label' => t('Fedora URI'),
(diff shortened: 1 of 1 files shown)


whikloj:
@kayakr I don't quite understanding when this appears from your description. You have saved the settings form but not selected any bundles?

kayakr:
@whikloj On a system upgrading to 8.x-1.0, where that settings form has never been saved. Possibly related to https://islandora.slack.com/archives/CM5PPAV28/p1568382357287900 also.

whikloj:
Ok, I think this just clicked. So I added this field but because you already had Islandora 8 installed it didn't get set via the [`config/install/islandora.settings.yml`](https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/config/install/islandora.settings.yml). Perhaps we should be including update hooks to set the default values.

PR #162, opened 2019-09-09
method check for remote videos

ajstanley:

Currently if we add a remote video media type Islandora whitescreens when it tries to use the non-existent referencedEntities() method.  This small change just checks for the method's existence before attempting to run.

# What's new?
One new check before execution.

# How should this be tested?
Add a remote video as a media. (You'll have to add mediaOf and mimeType as existing fields to the mediaType).

Try to display that media and you'll see a whitescreen.
Apply this pull, and you should see your embedded video.

# Interested parties
@Islandora-CLAW/committers


src/EventSubscriber/MediaLinkHeaderSubscriber.php
```diff
@@ -71,7 +71,8 @@ protected function generateMediaLinks(MediaInterface $media) {
     $source_field = $type_configuration['source_field'];
 
     if (empty($source_field) ||
-        !$media->hasField($source_field)
+        !$media->hasField($source_field ||
```
dannylamb: you're missing a closing parenthesis here before your `||`

ajstanley: It's three different conditions with logical ORs stuck between them,
if
  empty($source_field)
or
  !$media->hasField($source_field 
or 
  !method_exists($media->get($source_field), 'referencedEntities')

ajstanley: d'oh

dannylamb:
Just waiting on Travis...

PR #160, opened 2019-09-06
Hide media and members tabs on non-Islandora objects.

whikloj:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1057

# What does this Pull Request do?

Alters the route for the views to add a permission requirement and a custom function to check for 2 expected fields indicating its an Islandora object.

# What's new?

* Does this change require documentation to be updated? no
* Does this change add any new dependencies? no
* Does this change require any other modifications to be made to the repository
 (ie. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? no

# How should this be tested?

1. Get an islandora 8 instance running.
1. Create a Repository Item (assuming you are using the islandora_defaults module)
1. See the Media and Members tab when viewing the node as the Administrator
1. Create a Basic Page
1. See the Media and Members tab when viewing the node as the Administrator

1. Pull in this PR on top of the existing module.
1. Rebuild routes and clear caches
    `drupal --root=/var/www/html/drupal router:rebuild && drupal --root=/var/www/html/drupal cache:rebuild`
1. View the Repository item, see the Media and Members tabs
1. View the Basic page, **don't** set the Media and Members tabs

# Interested parties
@Islandora-CLAW/committers

PR #164, opened 2019-09-10
Work on #1270.

mjordan:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1270

* Other Relevant Links: https://github.com/Islandora-CLAW/islandora/pull/160#commitcomment-34979924

Adds 1) a Context Condition that checks to see if the current node has two fields: `field_model` and   `field_member_of` and 2) a Context Condition that checks to see if the current media has two fields, 'field_media_use' and 'field_media_of'.

# What's new?
* src/Plugin/Condition/NodeIsIslandoraObject.php
* src/Plugin/Condition/MediaIsIslandoraMedia.php

# How should this be tested?

## To test the NodeIsIslandoraObject condition:

1. Check out this PR.
1. Clear your cache (e.g. `drush cr`)
1. Create a new Context at `http://localhost:8000/admin/structure/context`
1. For a condition, use "Node is an Islandora object."
1. For a reaction, use "Blocks" (you could use any, but making a block show up is a good visual indicator the condition works as expected). Place the Devel block in the Footer, for example.
1. Save the Context.
1. View an islandora node. You should see the Devel links in your footer.
1. View a non-Islandora node (Article, Basic Page, etc.). You should not see the Devel links in your footer.
1. Edit your Context. Check "Negate the condition". Save the Context.
1. Revisit your two nodes. This time, the Devel links should show up on the non-Islandora node, and not show up on the Islandora node.

## To test the MediaIsIslandoraMedia condition:

1. Create a new media type by visiting `http://localhost:8000/admin/structure/media/add`. Choose whatever you want as a media source, e.g. 'File'.
1. After you have created this new Media type, create a media using this type. 
1. Check out this PR.
1. Clear your cache (e.g. `drush cr`)
1. Create a new Context at `http://localhost:8000/admin/structure/context`
1. For a condition, use "Media is an Islandora media."
1. For a reaction, use "Blocks" (you could use any, but making a block show up is a good visual indicator the condition works as expected). Place the Devel block in the Footer, for example.
1. Save the Context.
1. View an islandora media. You should see the Devel links in your footer.
1. View the non-Islandora media you created in step 2 You should not see the Devel links in your footer.
1. Edit your Context. Check "Negate the condition". Save the Context.
1. Revisit your two media. This time, the Devel links should show up on the non-Islandora media, and not show up on the Islandora media.

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/src/Plugin/Condition/NodeIsIslandoraObject.php
+++ b/src/Plugin/Condition/NodeIsIslandoraObject.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace Drupal\islandora\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\islandora\IslandoraUtils;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Checks whether node has fields that qualify it as an "Islandora" node.
+ *
+ * @Condition(
+ *   id = "node_is_islandora_object",
+ *   label = @Translation("Node is an Islandora object"),
+ *   context = {
+ *     "node" = @ContextDefinition("entity:node", required = TRUE , label = @Translation("node"))
+ *   }
+ * )
+ */
+class NodeIsIslandoraObject extends ConditionPluginBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * Islandora utils.
+   *
+   * @var \Drupal\islandora\IslandoraUtils
+   */
+  protected $utils;
+
+  /**
+   * Term storage.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManager
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructor.
+   *
+   * @param array $configuration
+   *   The plugin configuration, i.e. an array with configuration values keyed
+   *   by configuration option name. The special key 'context' may be used to
+   *   initialize the defined contexts by setting it to an array of context
+   *   values keyed by context names.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\islandora\IslandoraUtils $utils
+   *   Islandora utils.
+   * @param \Drupal\Core\Entity\EntityTypeManager $entity_type_manager
+   *   Entity type manager.
+   */
+  public function __construct(
+    array $configuration,
+    $plugin_id,
+    $plugin_definition,
+    IslandoraUtils $utils,
+    EntityTypeManager $entity_type_manager
+  ) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->utils = $utils;
+    $this->entityTypeManager = $entity_type_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('islandora.utils'),
+      $container->get('entity_type.manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $node = $this->getContextValue('node');
+    if (!$node) {
+      return FALSE;
+    }
+    // Islandora objects have these two fields.
+    if ($node->hasField('field_model') && $node->hasField('field_member_of')) {
+      return TRUE;
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    if (!empty($this->configuration['negate'])) {
+      return $this->t('The node is not an Islandora object.');
+    }
+    else {
+      return $this->t('The node is an Islandora object.');
+    }
+  }
+
+}
--- a/src/Plugin/Condition/NodeIsIslandoraObject.php
+++ b/src/Plugin/Condition/NodeIsIslandoraObject.php
@@ -12,7 +12,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  * @Condition(
  *   id = "node_is_islandora_object",
  *   label = @Translation("Node is an Islandora object"),
- *   context = {
+ *   context_definitions = {
  *     "node" = @ContextDefinition("entity:node", required = TRUE , label = @Translation("node"))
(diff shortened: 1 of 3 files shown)


src/Plugin/Condition/NodeIsIslandoraObject.php
```diff
@@ -0,0 +1,106 @@
...
+   *   Islandora utils.
+   * @param \Drupal\Core\Entity\EntityTypeManager $entity_type_manager
+   *   Entity type manager.
+   */
+  public function __construct(
+    array $configuration,
+    $plugin_id,
+    $plugin_definition,
+    IslandoraUtils $utils,
+    EntityTypeManager $entity_type_manager
```
whikloj: You don't seem to be using either the `IslandoraUtils` or `EntityTypeManager`, do you need them?

whikloj: Unless you are planning to test for non-node values of `$node = $this->getContextValue('node');` and load them.

mjordan: Nice catch @whikloj. That allowed me to remove the constructor too. I wasn't planning on testing for non-node values of `$node`.

whikloj:
@mjordan Just testing this I was wondering whether this context should also work for media as well as nodes?

src/Plugin/Condition/NodeIsIslandoraObject.php
```diff
@@ -0,0 +1,59 @@
...
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Checks whether node has fields that qualify it as an "Islandora" node.
+ *
+ * @Condition(
+ *   id = "node_is_islandora_object",
+ *   label = @Translation("Node is an Islandora object"),
+ *   context = {
```
whikloj: This key is now deprecated and should be changed to `context_definitions`
https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Condition!Annotation!Condition.php/property/Condition%3A%3Acontext/8.7.x

mjordan:
@whikloj do Islandora media have specific/required fields?

mjordan:
Or are they just attached to Islandora nodes?

whikloj:
I tried to do this work myself and it appears that while they do have a `field_media_of` they do not have a `field_model`. So perhaps another condition that is essentially `parentIsIslandoraNode`?

mjordan:
OK, I've added the MediaIsIslandoraMedia condition.

src/Plugin/Condition/NodeIsIslandoraObject.php
mjordan: Done in #2e5d865.

mjordan:
Bump :smiley:

rangel35:
I created a new content type for our needs and was going to ask if all I needed to do to make it an "Islandora" content type for context purposes was just adding those specific fields/tags

I am having problems getting the indexing to Fedora & Triplestore working so wanted to check to make sure I have all my bases covered

seth-shaw-unlv:
@rangel35 in addition to including `field_member_of` and `field_model` in your new content type, you also need to update the `repository_content` context condition ('admin/structure/context/repository_content') for it to index in Fedora and the triplestore.

whikloj:
I don't think we check for any fields right now when indexing content to Fedora. So updating the context condition as @seth-shaw-unlv says is probably all you need to do.

rangel35:
I just checked and I had updated the content type in the context before so I will make sure I have both fields and try again...thanks bunches

seth-shaw-unlv:
@rangel35 Just note that you may have to kick off indexing if this condition isn't in place before you create/update a node. 
Screen Shot 2019-09-25 at 12 41 18 PM

rangel35:
we made some changes to the contentType which will require purging content and re-ingesting so that will be a good test to see if it automatically does the index

dannylamb:
@rangel35 Also want to point out there's a lot of other ContextReactions going there doing altering to the RDF that you may find interesting.

rosiel:
SIDEBAR:
@dannylamb how many ContextReactions exist that alter the RDF, and how do we tell that they do that? 

I found that islandora.module calls any reaction that extends `NormalizerAlterReaction` - and it is the only module in [web root] that extends `NormalizerAlterReaction`:
- MappingUriPredicateReaction i.e. "Map URI to predicate" 
- JsonldTypeAlterReaction i.e. "Alter JSON-LD Type"

Based on their names, it's hard to tell what they do. Are there others? 

Controlled Access Terms also alters the RDF array, but it does some hardcoded transformations.

rangel35:
> we made some changes to the contentType which will require purging content and re-ingesting so that will be a good test to see if it automatically does the index

update:  I tested this on my local VM and it worked

whikloj:
@rosiel Currently those are the only two things "altering" json-ld that I am aware of.

side-sidebar: I am working on an example module to add a json-ld alter for my presentation at islandoracon. But essentially JSON-LD has a [hook](https://github.com/Islandora-CLAW/jsonld/blob/8.x-1.x/jsonld.api.php#L20) and islandora[ exposes that hook](https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/islandora.module#L248-L256) to the context framework so we can more easily add user configurable actions to do the altering.

dannylamb:
@rosiel In `islandora_defaults`, those two reactions are used in the 'All Media' and 'Content' contexts.

whikloj:
Lastly, because I failed to add it before.

* MappingUriPredicateReaction i.e. "Map URI to predicate"

  This defines the predicate which is used to add the currently rendered entities URI to the JSON-LD. Hence the default `schema:sameAs`. When this gets to Fedora then your Fedora object will have a `schema:sameAs <http://localhost:8000/node/X>` property.

* JsonldTypeAlterReaction i.e. "Alter JSON-LD Type"

   This adds the external URI for the content_model term into the JSON-LD as an rdf:type. So if you make an `Image`, which has an external_uri of `http://purl.org/coar/resource_type/c_c513`. Then this context reaction adds that URI as an `rdf:type`. You can test this, by viewing `http://localhost:8000/node/X?_format=jsonld`. Then removing this reaction from the "Content" context and view the json-ld again.

whikloj:
So this works as it is. 

However we have taken over the normal File and Image media types. So does that mean that a normal user should expect to have a type to upload images/files related to Islandora and one of other images/files? 

Otherwise (for media at least) we might need a better check...I'm not sure that that would be right now.

dannylamb:
Despite introducing a great deal more side questions, I think it's time to pull this in.  It's been reviewed and or tested or commented on by just about everyone ^_^

PR #166, opened 2019-09-12
IIIF Presentation API Validator

joecorall:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1263

Adds additional properties to IIIF manifest required by Presentation API

# What's new?
Adds a label to the manifest, and adds width/height to the canvas and image. Also adds the mimetype of the image. Since Drupal treats TIFFs differently than normal images, the height/width is calculated special for TIFFs using PHP's [getimagesize](https://www.php.net/manual/en/function.getimagesize.php)

# How should this be tested?

- Enable islandora_iiif submodule
- Add your IIIF Server URL (the same one you'd have configured for OSD) at /admin/config/islandora/iiif
- Import this "test" // "demo" view: https://gist.github.com/dannylamb/c90686957d3d5599f49f2cab02f045d4
- Access a node that has an image referencing the node via field_media_of e.g. /node/1
- Add "iiif-manifest" to the URL e.g. /node/1/iiif-manifest
- You should see the JSON of every image for the respective node
- Submit that URL to the IIIF Presentation API Validator at https://iiif.io/api/presentation/validator/service/

# Interested parties
@Islandora-CLAW/committers

### DIFF
--- a/modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
+++ b/modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
@@ -9,6 +9,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Serializer\SerializerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Drupal\Core\Config\ImmutableConfig;
+use Drupal\Core\File\FileSystem;
 
 /**
  * Provide serializer format for IIIF Manifest.
@@ -63,14 +64,22 @@ class IIIFManifest extends StylePluginBase {
   protected $iiifConfig;
 
   /**
+   * The Drupal Filesystem.
+   *
+   * @var \Drupal\Core\File\FileSystem
+   */
+  protected $fileSystem;
+
+  /**
    * {@inheritdoc}
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, SerializerInterface $serializer, Request $request, ImmutableConfig $iiif_config) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, SerializerInterface $serializer, Request $request, ImmutableConfig $iiif_config, FileSystem $file_system) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
 
     $this->serializer = $serializer;
     $this->request = $request;
     $this->iiifConfig = $iiif_config;
+    $this->fileSystem = $file_system;
   }
 
   /**
@@ -83,7 +92,8 @@ class IIIFManifest extends StylePluginBase {
       $plugin_definition,
       $container->get('serializer'),
       $container->get('request_stack')->getCurrentRequest(),
-      $container->get('config.factory')->get('islandora_iiif.settings')
+      $container->get('config.factory')->get('islandora_iiif.settings'),
+      $container->get('file_system')
     );
   }
 
@@ -105,12 +115,16 @@ class IIIFManifest extends StylePluginBase {
       $json += [
         '@type' => 'sc:Manifest',
         '@id' => $request_url,
+        // If the View has a title, set the View title as the manifest label.
+        'label' => $this->view->getTitle() ?: 'IIIF Manifest',
         '@context' => 'http://iiif.io/api/presentation/2/context.json',
         // @see https://iiif.io/api/presentation/2.1/#sequence
         'sequences' => [
-          '@context' => 'http://iiif.io/api/presentation/2/context.json',
-          '@id' => $iiif_base_id . '/sequence/normal',
-          '@type' => 'sc:Sequence',
+          [
+            '@context' => 'http://iiif.io/api/presentation/2/context.json',
+            '@id' => $iiif_base_id . '/sequence/normal',
+            '@type' => 'sc:Sequence',
+          ],
         ],
       ];
       // For each row in the View result.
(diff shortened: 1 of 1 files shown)


joecorall:
Do you have any error messages in `/admin/reports/dblog` or in your web server's log file?

rangel35:
thanks for the ideas...it was the view...once I imported Danny's view it worked...thanks to Eli Zoller for the help on slack...

joecorall:
Thanks! I updated the PR's instructions to use Danny's View.

rangel35:
encountered an issue...the manifest breaks if the top level has media.  all our books will have media on the "cover"

elizoller:
I was able to successfully validate the manifest with a TIF (file media) and Danny's view.

joecorall:
@rangel35 Are these the proper steps to reproduce the error?

1. Create a "book" islandora_object (let's call it `/node/1`)
2. Create a "page" islandora_object that has `field_member_of` set to `/node/1`
3. Add a TIFF file media to `/node/2` set to "Original File"
3. Add an image media to the `/node/1` set to "Thumbnail Image"
4. Access `/node/1/book-manifest` and a white screen error is displayed

rangel35:
yes that is what I was finding...it might not matter but but my order was...

Create a "book" islandora_object (let's call it /node/1)
Add an image media to the /node/1 set to "Thumbnail Image"
Create a "page" islandora_object that has field_member_of set to /node/1
Add a TIFF file media to /node/2 set to "Original File"
Access /node/1/book-manifest and a white screen error is displayed

elizoller:
Was also able to successfully validate the maniest with a JPG image (and with 1 jpg and 1 tif)

joecorall:
Thanks, @elizoller 

I can't seem to replicate the white screen issue. @rangel35 do you have any error messages in your web server logs or in `/admin/reports/dblog` that might give some clues?

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -157,32 +176,73 @@ protected function getTileSourceFromRow(ResultRow $row, $iiif_address, $iiif_bas
...
+            $width = $resource['width'];
+            $height = $resource['height'];
+          }
+          catch (ClientException $e) {
+          }
+          catch (ServerException $e) {
+          }
+
+          // If we couldn't get the info.json from IIIF
+          // try seeing if we can get it from Drupal.
```
whikloj: Just wondering if it might be faster to check Drupal first and fall back to the call to the IIIF server?

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
```diff
@@ -157,32 +176,73 @@ protected function getTileSourceFromRow(ResultRow $row, $iiif_address, $iiif_bas
...
+          if (empty($width) || empty($height)) {
+            // Get the image properties so we know the image width/height.
+            $properties = $image->getProperties();
+            $width = isset($properties['width']) ? $properties['width'] : 0;
+            $height = isset($properties['height']) ? $properties['height'] : 0;
+
+            // If this is a TIFF AND we don't know the width/height
+            // see if we can get the image size via PHP's core function.
+            if ($mime_type === 'image/tiff' && !$width || !$height) {
```
whikloj: Why only for a TIFF? Can this not work for other image formats as well?

joecorall: For other image formats, Drupal already calculated the width/height and stored it in the image properties above. Since TIFF is stored as a "File" media type, this isn't done so a special case was needed.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
joecorall: I debated this same concern with myself, definitely open to changing the logic around for speed concerns. I ended up going this route since:

1. Whatever is accessing this IIIF manifest is going to be accessing the same URL being accessed here, so if it's never been accessed before this will at least cache the response in Cantaloupe
2. It seemed cleaner to use IIIF for all image types (TIFF + everything else). The proceeding lines are essentially fallbacks incase IIIF isn't available.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
joecorall: Though I guess checking for TIFF isn't really necessary, is it? I think what you're getting at is if somehow Drupal failed to calculate the width/height, or it's not available, this could be an additional fallback to calculate on the fly for other formats.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
whikloj: Point taken, we will be serving the image from the IIIF server presumably right away. I'm fine leaving it in this order.

modules/islandora_iiif/src/Plugin/views/style/IIIFManifest.php
whikloj: 🤦‍♂ I always forget that TIFF is a file, we really need to see about getting it as an image. This is fine then, you could probably remove the check for `image/tiff` but I don't know it makes any difference in the end.

whikloj:
Guess I better get a publically accessible Islandora 8 instance up and running. 😈

whikloj:
@rangel35 are you are still experiencing the WSOD with the manifest? I had that problem but it was due fixed by https://github.com/Islandora-CLAW/islandora/pull/165. Can you check to ensure you have that commit or you will get the WSOD everytime.

Let me know how it works, I want to make sure I'm not missing something before I merge this.

rangel35:
@joecorall  @whikloj  I think #165 must have fixed my issue too...I rebuilt my vagrant and just tried the test and no more white screen...I get a manifest!!!

so consider my issue closed.

PR #167, opened 2019-09-17
Allowing NodeHasTerm to check for multiple tags using AND logic

dannylamb:
**GitHub Issue**: Part of https://github.com/Islandora-CLAW/CLAW/issues/932

Allows multiple terms to be used in the `NodeHasTerm`, `MediaHasTerm`, and `ParentNodeForMediaHasTerm` conditions.

# What's new?
A new `evaluate()` function in `NodeHasTerm`, which flows through to the other two condition plugins that extend it.

Plus I wrote a test.

# How should this be tested?

Try out the Condition in context.  Make it look for two terms.  Save the condition.  Then see if it works with nodes that do and do not have the terms.

# Interested parties
@Islandora-CLAW/committers @elizoller

### DIFF
--- a/src/Plugin/Condition/NodeHasTerm.php
+++ b/src/Plugin/Condition/NodeHasTerm.php
@@ -83,11 +83,12 @@ class NodeHasTerm extends ConditionPluginBase implements ContainerFactoryPluginI
    * {@inheritdoc}
    */
   public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $default = [];
     if (isset($this->configuration['uri']) && !empty($this->configuration['uri'])) {
-      $default = $this->utils->getTermForUri($this->configuration['uri']);
-    }
-    else {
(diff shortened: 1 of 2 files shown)


elizoller:
I can confirm that I am now able to add multiple terms separated by a comma and it applies only to nodes with both of the terms.

elizoller:
Except the term(s) keeps disappearing...

elizoller:
Nevermind, operator error.

seth-shaw-unlv:
This works as advertised. HOWEVER; the list of terms uses AND logic (i.e. all terms must be present) we should probably support a use-case where we can configure it to use OR logic (i.e. one or more terms present return TRUE). We would need to update the form to add a check-box and then add a conditional on the setting to choose between them.

@dannylamb should we merge as-is and add this as a future ticket; or would you rather add that in now?

PR #171, opened 2019-09-20
Integer-based weight drag-n-drop (Issue 1262)

seth-shaw-unlv:
**GitHub Issue**: Islandora-CLAW/CLAW/issues/1262

Test with Islandora-CLAW/islandora_defaults/pull/9

This PR ports the weight module's drag-n-drop code to work with our integer-based weight field (instead of using the weight module's weight field type). Yes, it means more code to support, _but_ that sacrifice means that we can have null-value weights (no weight value at all) for those repository items that don't need them and force positive values (which makes more sense to human-viewers of the resulting RDF).

# What's new?

* Added a new "Integer weight selector" with associated implementations of  hook_views_data_alter() and hook_preprocess_views_view_table().
* Does this change require documentation to be updated? We should add some documentation.
* Could this change impact execution of existing code? No.

# How should this be tested?

- Apply this PR.
- ~~Apply Islandora-CLAW/islandora_defaults/pull/9 and import the islandora_defaults feature.~~ _Doesn't actually matter anymore..._
- Make a parent item with some child objects that have weight values.
- (As a user with "Manage Members" privileges): visit the parent object record and click on the "Re-order Children" tab. Screen Shot 2019-09-16 at 7 30 32 AM
- Rearrange the children by grabbing the for-arrow icon and moving them around.  Screen Shot 2019-09-16 at 7 32 46 AM
- Alternatively, you can click the "Show row weights" link to manually select the weight values for each item.  Screen Shot 2019-09-16 at 7 37 38 AM
- Click "Save" when you are done. 
- Reload the page and the children should still be arranged in the new order.

# Additional Notes:

I made the selector work with any set of ordered items by making the possible position values match the ones found in the view. E.g. if you have 100 items and you are viewing items 81-90 then the re-order logic will only use those values for the re-ordering and won't mess with items above or below that range. (The weight module's selector would create a selector with _all_ possible range values which gets really big if we create very large ranges to accommodate big items.) This also means that if you want to move something to a different range (e.g. from position 80 to position 8) you either need to adjust the "items per page" being displayed or first manually edit the child's weight on it's edit tab.

# Interested parties
@dannylamb, @Islandora-CLAW/committers

### DIFF
--- a/islandora.links.action.yml
+++ b/islandora.links.action.yml
@@ -9,3 +9,9 @@ islandora.add_member_to_node:
   title: Add child
   appears_on:
     - view.manage_members.page_1
+
+islandora.reorder_children:
+  route_name: view.reorder_children.page_1
+  title: Reorder Children
+  appears_on:
+    - view.manage_members.page_1
--- a/islandora.module
+++ b/islandora.module
@@ -409,3 +409,47 @@ function islandora_entity_view(array &$build, EntityInterface $entity, EntityVie
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_views_view_table().
+ *
+ * Used for the integer-weight drag-n-drop. Taken almost
+ * verbatim from the weight module.
+ */
+function islandora_preprocess_views_view_table(&$variables) {
+
+  // Check for a weight selector field.
+  foreach ($variables['view']->field as $field_key => $field) {
+    if ($field->options['plugin_id'] == 'integer_weight_selector') {
+
+      // Check if the weight selector is on the first column.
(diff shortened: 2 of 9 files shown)


seth-shaw-unlv:
@dannylamb this is ready for testing. Travis is still going, but already passed the test I added.

dannylamb:
Thank you for the tests @seth-shaw-unlv.  Even if they are straight ripped from core, it adds to your positive karma :prayer_beads:

seth-shaw-unlv:
@dannylamb well, ripped from a contrib module, at least.

dannylamb:
Just waiting on Travis.   And thanks again for the tests :+1: :+1: :+1:

PR #174, opened 2019-09-30
Make fedora uri available on taxonomy terms (issue-1288)

seth-shaw-unlv:
**GitHub Issue**: Islandora-CLAW/CLAW/issues/1288

Fixes a route parameter check to work for taxonomy terms and not just nodes and media.

# What's new?

* Changes islandora_entity_view from using getParameters to getRawParameters and update the check to see if the entity type is in the list of parameter keys.
* Does this change require documentation to be updated? No.
* Could this change impact execution of existing code? Fixes it.

# How should this be tested?

- Adjust your islandora settings to display the Fedora URI on some taxonomy vocabularies.
- Visit one of them and see no Fedora URI 😞
- Apply the PR
- Clear caches
- Visit the term again and see the Fedora URI 🎉

# Additional Notes:
Any additional information that you think would be helpful when reviewing this
 PR.

# Interested parties
@Islandora-CLAW/committers, esp. @dannylamb and @rosiel

### DIFF
--- a/islandora.module
+++ b/islandora.module
@@ -379,7 +379,7 @@ function islandora_entity_extra_field_info() {
  * Implements hook_entity_view().
  */
 function islandora_entity_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
-  $route_match_item = \Drupal::routeMatch()->getParameters()->all();
+  $route_match_item = \Drupal::routeMatch()->getRawParameters()->all();
   // Ensure the entity matches the route
   if (array_key_exists($entity->getEntityTypeId(), $route_match_item)) {
     if ($display->getComponent('field_gemini_uri')) {
--- a/islandora.module
+++ b/islandora.module
@@ -380,10 +380,8 @@ function islandora_entity_extra_field_info() {
  */
 function islandora_entity_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
   $route_match_item = \Drupal::routeMatch()->getParameters()->all();
-  // Get the parameter, which might be node, media or taxonomy term.
-  $current_entity = reset($route_match_item);
-  // Match exactly to ensure they are the same entity type too.
-  if ($entity === $current_entity) {
+  // Ensure the entity matches the route
+  if (array_key_exists($entity->getEntityTypeId(), $route_match_item)) {
     if ($display->getComponent('field_gemini_uri')) {
       $gemini = \Drupal::service('islandora.gemini.lookup');
       if ($gemini instanceof GeminiLookup) {


dannylamb:
Testing now.

dannylamb:
Nice.  Switching to raw params normalizes everything out. I can confirm this catches all three entity types: node, media, and taxonomy terms.  :+1:

Just waiting on Travis.

seth-shaw-unlv:
Actually the fix works even without the switch to raw parameters as it focuses on using the the parameter array keys rather than trying to load the entities and matching them.

So really, now that I consider it again, I've just made that first check more permissive as we aren't forcing entity equivalency.

seth-shaw-unlv:
Oh Travis, why must you fail on breadcrumbs? I didn't change anything there....

seth-shaw-unlv:
@dannylamb, I decided to change tactics to tighten up the check based on [my earlier comment](#issuecomment-536663128). So, it should be re-tested.

I'll investigate why Travis died on breadcrumbs.

islandora.module
```diff
@@ -379,11 +379,9 @@ function islandora_entity_extra_field_info() {
  * Implements hook_entity_view().
  */
 function islandora_entity_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
-  $route_match_item = \Drupal::routeMatch()->getParameters()->all();
-  // Get the parameter, which might be node, media or taxonomy term.
-  $current_entity = reset($route_match_item);
-  // Match exactly to ensure they are the same entity type too.
-  if ($entity === $current_entity) {
+  $route_match_item = \Drupal::routeMatch()->getParameters()->get($entity->getEntityTypeId());
```
whikloj: So this view is called for every object rendered in a page.

The first route match seems to be the main entity which is why I pulled it off the top the route. That check was to avoid displaying the Fedora URI for both the node and the media entity on the same page.

Perhaps we need a better way to avoid that, but I think this is going to display them all (them == every entity that has a Fedora URI embedded on this page) on every page they are rendered on.

whikloj: More details about why I added this in like this.
https://github.com/Islandora-CLAW/islandora/pull/129

seth-shaw-unlv: Well, that is easy enough to test. If we have a collection displaying the members block it should only show the URI for the collection and not the members... that seems to be what is happening. I'm not seeing any issues surface with the current commit.

seth-shaw-unlv:
Of course, it died on coding standards, not the tests! At least those are easy to fix!

islandora.module
whikloj: This doesn't seem to break when displaying the view of Children because it doesn't get called by the children view. So as long as people test before embedding a node in a node or media in a media I think this is fine.

seth-shaw-unlv: So, should we be running this by the folks using Paragraphs? I can't recall who was doing that...

whikloj: Maybe, this is one of those what is "your normal". Which for some might be many levels of inception deep. Me, I'm shallow like a puddle.

Paging @alxp @ajstanley for Paragraphs expertise.

seth-shaw-unlv: Honestly, shallow works for me; which is why I'm happy with this solution as-is. I just don't want us to merge this to fix one issue and create another at the same time.

PR #175, opened 2019-10-01
Issue 1291

ajstanley:
**GitHub Issue**: (https://github.com/Islandora-CLAW/CLAW/issues/1291)

Odd failures will occur is a user who has not been assigned the hardcoded fedoraAdmin role
attempts to add a media type that is set to save to the fedora filesystem.

# What's new?
This fix adds a filter to the controller so the user will not see media types they cannot use.

# How should this be tested?
Create a user with the Administrator role, but without fedoraAdmin role.
Log in or masquerade as that user.
Have at least one media type that saves to fedora and one that does not.
(This setting is found under the field settings for the file field in the Media type).
Navigate to the Add Media button on the Media tab of any object.
You should only see media types that do NOT save to fedora.

# Interested parties
 @Islandora-CLAW/committers


whikloj:
Looks good @ajstanley.

You can inject the current user in this controller or the [ManageMembers](https://github.com/Islandora-CLAW/islandora/blob/8.x-1.x/src/Controller/ManageMembersController.php#L49) one it extends using something like this (https://api.drupal.org/comment/62807#comment-62807)

whikloj:
Also, just thinking. This problem is due to not having a WebAC acl to allow `fedoraUser`s the "write" permission. So eventually we will want to remove this check if we have setup allowed roles in Fedora. Then permissions could be defined by any of the other roles from Drupal which are passed into Fedora.

Not saying that this isn't valid, just that I don't think it will always be needed unless we are thinking that we will be allowing all users administrator write access to our Fedora.

seth-shaw-unlv:
I'm concerned that this will block (or at least make it more difficult) to add non-Fedora synced media to a site. E.g. we have a page with a photo of the repository, but we can't add the photo because the Image media is a "Fedora" image. Instead of filtering out these Media that we've configured to use the fedora storage scheme _by default_, why don't we try adding a file scheme to the Media creation form and filter out the fedora option there if they don't have the fedoraAdmin role?

ajstanley:
@whikloj  - Good suggestion - class property added at the MediaManager level in case this ever gets extended again and someone else needs user data.

I've no idea how locked down the various roles and role names are.  Maybe it would be worth making roles with fedora access definable in the Islandora settings?

whikloj:
@seth-shaw-unlv I haven't tested this yet to see what it ends up looking like. I feel like this is a solution for the short term as ideally I don't want to give all users `fedoraAdmin` rights to create content in my Fedora. But I think that is going to require some work on the Fedora/WebAC side define how we want that to look.

@ajstanley currently the Syn tomcat valve looks for the role `fedoraAdmin` to give admin rights. It would probably make sense to make that configurable in the `syn-settings.xml` then you could choose the role name in Drupal you want (like `administrator`).

rosiel:
@seth-shaw-unlv In your hypothetical, there's a user who understands the differences between media located in fedora and located locally. They have opinions on where things should be saved. Right now, unless they can add a media type, they can't save anything to the drupal file system (because in islandora/islandora_defaults, the existing drupal media types have been hijacked to point to Fedora). What you're describing seems like kind of a different issue - to what extent, in the default implementation, should a user be allowed to choose whether a media file should live in Fedora or Drupal? 

Who should have an understanding of WebAC, access to configure Syn, and manage drupal roles? Can we add docs for that person in the github docs?

seth-shaw-unlv:
I played around with this a bit. Yes, I can have a non-fedoraAdmin user create an article and upload an image as a File entity and links it in article's markup just fine. But this bypasses Drupal media altogether. 

However, our library is using the new Media Library for re-usable content. We add media to the library using the same form which leads to the various errors if someone doesn't have the fedoraAdmin role. We want to use the media library for things like logos and stock images; but we don't want that content in our fedora repository. So, even if we use this PR to make the UI better going from the Media tab, we still need to address this other issue.

So, yes, a user in our library _will_ need to understand the difference between an Image Media in Fedora and one in Drupal if we want to use the Media library as it was intended.

src/Controller/ManageMediaController.php
```diff
@@ -22,14 +22,26 @@ class ManageMediaController extends ManageMembersController {
...
       'entity.media.add_form',
       'entity.media_type.add_form',
       $node,
       'field_media_of'
     );
+    if (!in_array('fedoraadmin', $roles)) {
+      $bundles = $list['#bundles'];
+      foreach ($bundles as $label => $bundle) {
+        $storage = $this->entityFieldManager->getFieldStorageDefinitions('media', $label);
+        $scheme = $storage['field_media_file']->getSetting('uri_scheme');
```
rosiel: Only the 'File' Media type contains the `field_media_file` field. 
'Audio' contains `field_media_audio_file`.
'Image' contains `field_media_image`.
'Video' contains `field_media_video_file`.

ajstanley: Nice catch @rosiel - That method looked like it was filtering, but it wasn't.  New logic is in place and tested against a couple of file types.

@seth-shaw-unlv This should handle your concern.  You can have two media types ie. image_fedora and image_drupal and use them depending on where you want the file to live. It's not possible to have the same media type stored in two different places - Drupal won't allow it.

seth-shaw-unlv:
@ajstanley you could have two types, but if that is the case then Islandora should be making that the default rather than modifying the original.

It _is_ possible to have a Media type store items in multiple filesystem schemes. Islandora 8 already does. E.g. http://future.islandora.ca/index.php/en/media/37 is an 'Original File' that is stored in Fedora (note the `_flysystem` in http://future.islandora.ca/index.php/_flysystem/fedora/2019-06/animal-cat-cute-46024.jpg ) while it's 'Service File' derivative http://future.islandora.ca/index.php/en/media/38 is stored on the Drupal Public file system (note the `sites/default/files` in http://future.islandora.ca/sites/default/files/2019-06/18-Service%20File.jpg).

ajstanley:
@seth-shaw-unlv - Excellent point.  In this case, the original file is stored according to the rules set up in the image media type, and the service file gets its file system from the Action configuration.  The fields with types of image and file differ (oddly) in their behavior.  Both claim that the file storage location cannot be changed in the field definition if there is data in the database, but for images, you can ignore the warning and change it anyway,
[Image field storage](http://future.islandora.ca/index.php/en/admin/structure/media/manage/image/fields/media.image.field_media_image/storage)
and for files
[File field storage](http://future.islandora.ca/index.php/en/admin/structure/media/manage/file/fields/media.file.field_media_file/storage)
You cannot.
I don't know of any way of making the storage location user-selectable.  How are you doing this now?

seth-shaw-unlv:
Despite the field storage setting, the only place the scheme is stored is on the file entity it references' uri property. So, the only time the field storage setting actually matters is for the web-form upload. Using the Migrate API? Use either 'fedora://' or 'public://' on the file entity's uri value. Using islandora_workbench? Configure it to do the same on the REST PUT requests.

For most Drupal sites hiding this choice between public and private filesystems (i.e. 'always available' v. 'I can control who has access') makes good UI sense because it is an early architecture decision and not one to be decided on a case-by-case basis. Adding in this additional filesystem scheme that we may or may not want to use for a given media complicates this and, necessarily, requires more choices.

We can avoid the whole issue by just assuming everything goes into the repository, using different loading mechanisms (what we've done to date), making new Media types (as was proposed), or make the Media upload forms support choosing between available filesystems. My preference would be the last option, if possible. 

To that end, we should be able to extend the existing upload widgets (e.g. FileWidget), but I have not tested that theory.

ajstanley:
I gotta admit it's a use case I hadn't encountered (or even considered) before, but it's an interesting one.  This solution would still work with a bespoke widget as long as the file storage uri_scheme didn't point to fedora.  If the proposed widget was to be a part of core Islandora it would be pretty easy to filter on the presence of that filter in any Media type's form.
Yet another solution, which would involve more work probably than necessary, would be to have a whole new field type.  This filter triggers on $field->getType() and only reacts to file and image.

PR #177, opened 2019-10-11
Use a more illustrative name for the "Drupal uri predicate" / "Map URI to predicate" reaction.

rosiel:
**GitHub Issue**: https://github.com/Islandora-CLAW/CLAW/issues/1303

To channel @seth-shaw-unlv , 🎶 If I go there's just no telling how far I'll go 🎶

*  Per Eli's and my diagram, this is how we put (excuse the lazy rdf) `<node/1> schema:sameAs <node/1>` in the JSON-LD of node/1,  in order to have `<fedoraURI> schema:sameAs <node/1>`. The schema:sameAs is the "Drupal URI Predicate" in the "Map Drupal URI to Predicate" reaction. Those labels, alone, do not explain what's going on.

Renames almost all uses of "Map URI to predicate"-type wording into "JSON-LD Self-Reference Predicate"-type wording.

# What's new?

Front-end: 

* Renames the Reaction's human-readable label to "JSON-LD self-reference".
* Renames the human-readable label variable in the config form to 'Self-reference predicate'
* Renames the human-readable description of that variable in the config form to "When creating the JSON-LD for this Drupal entity, add a relationship from the entity to itself using this predicate. It must use a defined RDF namespace prefix."
* renames the human-readable summary for this reaction (! where does this appear??) to 'When creating the JSON-LD for this Drupal entity, add a relationship to itself using this predicate.'

Back-end:

* renames the Reaction's class to `JsonldSelfReferenceReaction`, and renames the php file.
* renames the constant to `SELF_REFERENCE_PREDICATE`
* renames some internal variables
* renames the test files to match, and to find the correct config form field.

Does not change:

* the internal id of the reaction remains `islandora_map_uri_predicate`

Could this change impact execution of existing code?

* I don't think so. I think that because the internal id didn't change, this should just work (after a cache clear)

# How should this be tested?

* Pull this code into islandora
* *Clear your cache*
* Look at existing Contexts that contain this reaction (e.g. Content). The wording above should appear in the interface.
* the context should continue to work as intended, i.e. this causes a "schema:sameAs" to appear in the json-ld of the node, pointing to itself.
* you can add this reaction to another context and it works as intended. 
* best way to test: show this to someone who doesn't know how this works and ask them what they think it does.

# Additional Notes:
...

# Interested parties


dannylamb:
Closing since https://github.com/Islandora/islandora/pull/729 has been merged

PR #179, opened 2019-12-12
Add file to derivative

ajstanley:

**Don't Merge yet!!**
This is the work we've done on an Islandora Model which allows derivatives to be created and attached to existing media rather than creating an entire new Media.

[Link to some documentation](https://docs.google.com/document/d/1nnczfKToQq5qIYrD9x6MKH9vSApgK6lNdK3PHCHdmSU/edit)

# What's new?
A new Abstract Class to create the Actions
A new ContextReaction to deal with them
An example Action in the Islandora Image module
A New endpoint to get the derivative file from Karaf to attach to the media.

This does not affect any current Islandora behaviour.

# How should this be tested?

- Load up this Islandora,
- Add some fields into your Media Types to hold the returned files
- Create some actions
- Create some new contexts
- Ingest some Media objects.

# Additional Notes:
I'll put together a features module with some sensible defaults.

# Interested parties
 @Islandora-CLAW/committers


dannylamb:
This should definitely be generalized and made against the core islandora module.  I see very little that's OCR specific here...  but wrong fork.  This one is kept around until a handful of other issues get taken care of, then it goes away.  Make the PR against https://github.com/Islandora/islandora

